PR #3, opened 2021-01-16
add IsAdmin component

fabiankaegy:
This is a component I had to use several times in projects. It allows the developer to check wether the current logged in user is an admin and render items only if it is the case. 

An example for the use case: 
```js
import { IsAdmin } from '@10up/block-components';
function MyComponent( props ) {
    return (
        <IsAdmin
            fallback={ <p>Sorry, you are not allowed to do that</p> }
        >
            <p>Only Administrators can see what you put in here</p>
        </IsAdmin>
    )
}
```

This very much mirrors the [`MediaUploadCheck`](https://github.com/WordPress/gutenberg/blob/004a32ec09d5b789be694d85725fe49cca385f08/packages/block-editor/src/components/media-upload/README.md#usage) component from Gutenberg directly.


ryanwelcher:
@fabiankaegy code looks good to me.

I wonder if we can add a more concrete example? I can see users not being totally clear on why/when this component should be used.

At a high-level, I think we should be overly verbose on use-cases and details for these components - especially ones that may not be used extensively or have nice use-cases.

fabiankaegy:
@ryanwelcher Yeah that is true. It is a very special use case that should be described more concise :) Will take a look at a better example

fabiankaegy:
@ryanwelcher I have added some more context to the readme and linked to a real world example. A version of this component was actually build in the Maps Block Apple :)

PR #5, opened 2021-01-18
Changes to the package.json.

ryanwelcher:
Changes to the package.json file to prep for eventual publishing to NPM

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,10 @@
 {
-  "name": "10up-block-components",
+  "name": "@10up/block-components",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -15,8 +18,12 @@
     "WordPress",
     "block-editor"
   ],
-  "author": "Faian Kägy",
```
fabiankaegy, line 18: :O Very good catch

fabiankaegy (approved):
Looks good to me  👍

PR #6, opened 2021-01-20
[WIP] A bunch of clean up on content picker

tlovett1:
Preparing content picker for using terms as well as posts


components/ContentPicker/SortableList.js
```diff
@@ -0,0 +1,49 @@
+import { SortableContainer, SortableElement } from 'react-sortable-hoc';
+import PickedItem from './PickedItem';
+
+const { useSelect } = wp.data;
```
fabiankaegy: @tlovett1 We need to use the import syntax for all wordpress dependencies for the dependency extraction plugin to pull it in correctly.

tlovett1: Didn't know that. Thanks!

PR #7, opened 2021-01-29
Add: `Image`, `MediaToolbar` Components and `useMedia` hook

fabiankaegy:
## `Image` Component
The `Image` component is built to make it easier to work with Images in the Block Editor. The component only needs the image id and will pull the complete image via the Data API and create the markup for it. It has support to specify the `size` of the image so it becomes easier to use the same image dimensions that the frontend uses.

## `MediaToolbar` Component
The `MediaToolbar` component is built to quickly add a complete media editing experience to the `BlockControls` toolbar. It shows an "Add Image" or "Replace Image" + "Remove Image" buttons that handle the full flow. Whether or not an image can be removed can be configured using the `isOptional` prop. 

### Usage
```js
import { MediaToolbar } from '@10up/block-components';

function BlockEdit(props) {

    const { attributes: { imageId }, setAttributes } = props;

    function handleImageSelect( image ) {
        setAttributes({imageId: image.id});
    }

    function handleImageRemove() {
        setAttributes({imageId: null})
    }

    return (
        <>
            <BlockControls>
                <MediaToolbar
                    isOptional
                    id={ image }
                    onSelect={ handleImageSelect }
                    onRemove={ handleImageRemove }
                />
            </BlockControls>
            ...
        </>
    )
}
```

## `useMedia` hook
The `useMedia` hook accepts an ID and returns the full image object from WordPress. It makes it easy to only store the ID in the attributes and retrieve the rest live in the editor same as we do in PHP.


tlovett1:
What is `image` supposed to look like?

ryanwelcher:
I am not sold on storing the whole image object as an attribute, it's gigantic. From a performance perspective it has to have an impact if we have to parse the entire object just to retrieve the URL to show the image. To be clear, this is not specific to this component, but just a general issue we need to solve for.

One approach I have tried is to store only the media ID and then retrieve the media via the datastore when block is loaded and store that object in internal state. There are consideration such as loading state and whether the images is loading on mount or on select however. 

Sorry if this derailing this a bit but I think we need to be thoughtful here before releasing an approach 😄

tlovett1:
@fabiankaegy any thoughts on Ryan's feedback?

fabiankaegy:
@tlovett1 I have refactored this so it no longer stores all the info in the attributes. It now stores only the ID and the rest is retrieved using the data api.

fabiankaegy:
@tlovett1 This component is ready for final review 👍

tlovett1 (approved):

PR #9, opened 2021-02-03
refactor useHasSelectedInnerBlock to use core function

fabiankaegy:
I just found out that core now has a method in the `core/block-editor` data store to determine wether a block has block selected. So I refactored our component to use the core functionality.


tlovett1:
Nice!

Issue #18, opened 2021-03-18
Content Picker search can run less often


johnwatkins0:
**Is your enhancement related to a problem? Please describe.**
The search runs every time input changes. While this is generally not a problem, it can be an issue on slow connections, and it is technically possible to type so fast that you reach the limit of the simultaneous requests the browser will allow.

**Describe the solution you'd like**

No need to debounce. Instead, we can just make sure only one fetch request is running at a time. After a fetch completes, we check whether the search term has changed; if it has, search again. This way we will still make a lot of requests that are never surfaced in the UI as the user types, but it will be fewer requests than what the component does now.

**Designs**

**Describe alternatives you've considered**

Debouncing, though as described above, I don't think it's necessary.

**Additional context**

I'll send a PR for this, unless someone else sees this and wants to try it.

xavortm:
Also probably fixed in #71

fabiankaegy:
Closing this as it appears to be fixed as of #71

PR #19, opened 2021-03-18
Fix/content picker improvements

tlovett1:
### Description of the Change

* Changes class names to make more sense for the Content Picker/Search
* Remove type label from search results if only one type is being searched.
* Caching for content search

### Possible Drawbacks

* Backwards compat break.

Addresses #18

### DIFF
--- a/components/ContentSearch/SearchItem.js
+++ b/components/ContentSearch/SearchItem.js
@@ -15,6 +15,8 @@ import { Button, TextHighlight } from '@wordpress/components';
  * @return {*} React JSX
  */
 const SearchItem = ({ suggestion, onClick, searchTerm, isSelected, id, contentTypes }) => {
+	const showType = suggestion.type && contentTypes.length > 1;
+
 	return (
 		<Button
 			id={id}
@@ -27,14 +29,25 @@ const SearchItem = ({ suggestion, onClick, searchTerm, isSelected, id, contentTy
 			}}
 		>
 			<span className="block-editor-link-control__search-item-header">
-				<span className="block-editor-link-control__search-item-title">
+				<span
+					className="block-editor-link-control__search-item-title"
+					style={{
+						paddingRight: !showType ? 0 : null,
+					}}
+				>
 					<TextHighlight text={decodeEntities(suggestion.title)} highlight={searchTerm} />
 				</span>
-				<span aria-hidden className="block-editor-link-control__search-item-info">
+				<span
+					aria-hidden
+					className="block-editor-link-control__search-item-info"
+					style={{
+						paddingRight: !showType ? 0 : null,
+					}}
+				>
 					{filterURLForDisplay(safeDecodeURI(suggestion.url)) || ''}
 				</span>
 			</span>
-			{suggestion.type && contentTypes.length > 1 && (
+			{showType && (
 				<span className="block-editor-link-control__search-item-type">
 					{/* Rename 'post_tag' to 'tag'. Ideally, the API would return the localised CPT or taxonomy label. */}
 					{suggestion.type === 'post_tag' ? 'tag' : suggestion.type}
--- a/components/ContentSearch/SearchItem.js
+++ b/components/ContentSearch/SearchItem.js
@@ -14,7 +14,7 @@ import { Button, TextHighlight } from '@wordpress/components';
  * @param props.id
  * @return {*} React JSX
  */
-const SearchItem = ({ suggestion, onClick, searchTerm, isSelected, id }) => {
+const SearchItem = ({ suggestion, onClick, searchTerm, isSelected, id, contentTypes }) => {
 	return (
 		<Button
 			id={id}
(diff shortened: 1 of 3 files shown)


components/ContentSearch/index.js
```diff
@@ -5,7 +5,9 @@ import PropTypes from 'prop-types';
 import { __ } from '@wordpress/i18n';
 import SearchItem from './SearchItem';
 
-const NAMESPACE = '10up-block-components';
+const NAMESPACE = '10up-content-search';
+
+const searchCache = {};
```
johnwatkins0, line 10: This should be state within the component, because if it's at the module level, then multiple instances of the component will end up using the same cache object, even though they may have different contentTypes. Alternatively, we could key this cache object by the full search URL (or JSON.stringified query args) instead of just the keyword. Then it wouldn't have to be state.

components/ContentSearch/index.js
```diff
@@ -51,30 +53,37 @@ const ContentSearch = ({ onSelectItem, placeholder, label, contentTypes, mode, e
 		setSearchString(keyword);
```
johnwatkins0: This may look like overkill and is totally optional for this PR, but once the search cache is converted to state, we can handle search changes a little more efficiently (and close #18) by responding to keyword changes with effect hooks.

```JS
/**
	 * handleSearchStringChange
	 *
	 * Using the keyword and the list of tags that are linked to the parent block
	 * search for posts/terms that match and return them to the autocomplete component.
	 *
	 * @param {string} keyword search query string
	 */
	const handleSearchStringChange = (keyword) => {
		setSearchString(keyword);
	}

	useEffect(() => {
		if (isLoading || searchString in searchCache) {
			return;
		}

		setIsLoading(true);

		const searchQuery = `wp/v2/search/?search=${searchString}&subtype=${contentTypes.join(
			',',
		)}&type=${mode}`;

		apiFetch({
			path: searchQuery,
		}).then((results) => {
			const newResults = results.filter((result) => {
				let keep = true;

				if (excludeItems.length) {
					excludeItems.forEach((item) => {
						if (item.id === result.id) {
							keep = false;
						}
					});
				}

				return keep;
			});

			setSearchCache(oldSearchCache => ({...oldSearchCache, [searchString]: newResults}));
			setIsLoading(false);
		});
	}, [isLoading, searchCache, searchString]);

	useEffect(() => {
		if (searchString in searchCache) {
			setSearchResults(searchCache[searchString])
		}
	}, [searchString, searchCache])
```

The first effect hook will run when the search term changes, when the cache changes, and when `isLoading` changes. This includes when `isLoading` becomes false -- guaranteeing that the latest search term will always be queried. But by having that `isLoading` check at the start of the effect hook, we also ensure that only one fetch request is running   at a time.

The second effect hook responds to changes to the search cache. This will run after a successful `fetch` to populate the results.

components/ContentSearch/index.js
```diff
@@ -51,30 +53,37 @@ const ContentSearch = ({ onSelectItem, placeholder, label, contentTypes, mode, e
...
+		} else {
+			const searchQuery = `wp/v2/search/?search=${keyword}&subtype=${contentTypes.join(
+				',',
+			)}&type=${mode}`;
+
+			apiFetch({
+				path: searchQuery,
+			}).then((results) => {
+				const newResults = results.filter((result) => {
+					let keep = true;
```
johnwatkins0, line 74: Since this is async, we should bail here if the component has unmounted. This can be achieved with `useRef` and an effect hook somewhere at the top of the component.

```JS
const mounted = useRef(true);

	useEffect(() => {
		return () => {
			mounted.current = false;
		};
	}, []);
```

That effect hook's return callback will set `mounted` to false when the component is destroyed. Then if `mounted.current` is false, we can bail out of the `fetch` callback before setting state.

components/ContentSearch/SearchItem.js
```diff
@@ -14,7 +14,9 @@ import { Button, TextHighlight } from '@wordpress/components';
  * @param props.id
  * @return {*} React JSX
  */
-const SearchItem = ({ suggestion, onClick, searchTerm, isSelected, id }) => {
+const SearchItem = ({ suggestion, onClick, searchTerm, isSelected, id, contentTypes }) => {
+	const showType = suggestion.type && contentTypes.length > 1;
```
johnwatkins0, line 18: This is working well.

johnwatkins0 (changes requested):
@tlovett1 The changes are working well, but I couldn't resist a few higher-level suggestions. Feel free to ignore those as they can be tackled separately. The only one I would call required is moving the search cache into state for the reason described in my inline comment.

components/ContentSearch/index.js
```diff
@@ -51,30 +53,37 @@ const ContentSearch = ({ onSelectItem, placeholder, label, contentTypes, mode, e
...
+					if (excludeItems.length) {
+						excludeItems.forEach((item) => {
+							if (item.id === result.id) {
+								keep = false;
+							}
+						});
+					}
+
+					return keep;
+				});
```
johnwatkins0, line 85: At some point we should handle a fetch error here.

tlovett1:
Great feedback @johnwatkins0 . I pushed up some changes. I left the search cache outside of the component because I think it's really important to share search cache across multiple instances.

johnwatkins0 (approved):
Looks good.

PR #32, opened 2021-06-11
Content picker: handle nonexistent items

### LINKED ISSUE #26: Content Picker: deleted posts not handled
**Describe the bug**

With the ContentPicker, we need a way to let the top-level component know when one or more of the previously selected posts has been deleted.

**Steps to Reproduce**

1. Set up a Content Picker component with the picked posts being saved to post meta.
2. Set maxContentItems to 3
3. Select three posts with the Content Picker and save
4. Delete two of the posts
5. Reload.
6. The content picker only renders the non-deleted posts, but it still thinks there are three posts (length of the `content` prop is 3), so it doesn't render the search bar or provide a way to remove the deleted posts.

**Expected behavior**

ContentPicker should find out when its content can't be retrieved and update its content prop accordingly.

**Screenshots**

**Environment information**
 - Device: 
 - OS: 
 - Browser and version: 
 - WordPress version: 
 
 - Plugins and version: 
 - Theme and version: 
 
 - <details><summary>Site Health Info:</summary>
   
   </details>

**Additional context**

johnwatkins0:
### Description of the Change

This resolves #26 by running the delete callback on any items that are not returned by `getEntityRecord`. I also did a little tangential cleanup, which I'll comment on inline.

### Alternate Designs

This change only applies to items that have been deleted completely. So e.g. draft posts or posts that are in trash are still returned. We might want to do something to handle those cases, maybe in a separate issue.

### Benefits

When a previously picked item is nonexistent, the content picker will no longer behave as if there are more items than there really are.

### Verification Process

When using this branch, use the content picker in a context where its picked items are saved and loaded from the database (e.g., in post meta). Pick an item and save. Delete the item completely. Reload the content picker and verify that the `maxContentItems` works as expected.

### Checklist:

### Applicable Issues

#26

### Changelog Entry

Add handling of nonexistent items in the ContentPicker component.

### DIFF
--- a/components/ContentPicker/index.js
+++ b/components/ContentPicker/index.js
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types';
 import arrayMove from 'array-move';
 import styled from '@emotion/styled';
 import { select } from '@wordpress/data';
-import { useState } from '@wordpress/element'; // eslint-disable-line
+import { useState, useEffect, useMemo } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import { ContentSearch } from '../ContentSearch';
 import SortableList from './SortableList';
@@ -10,9 +10,24 @@ import SortableList from './SortableList';
 const NAMESPACE = '10up-content-picker';
 
 /**
+ * Unfortunately, we had to use !important because on PickedItem we couldn't @emotion/styled css
+ * as it was breaking sortability from react-sortable-hoc
+ */
+const StyleWrapper = styled('div')`
+	& .block-editor-link-control__search-item {
+		border: none !important;
+		cursor: default;
+
+		&:hover {
+			background: transparent;
+		}
+	}
+`;
+
+/**
  * Content Picker
  *
- * @param {Object} props React props
+ * @param {object} props React props
  * @param props.label
  * @param props.mode
  * @param props.contentTypes
@@ -25,7 +40,7 @@ const NAMESPACE = '10up-content-picker';
  * @param props.content
  * @param props.uniqueContentItems
  * @param props.excludeCurrentPost
- * @return {*} React JSX
+ * @returns {*} React JSX
  */
 const ContentPicker = ({
 	label,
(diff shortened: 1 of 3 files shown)


components/ContentPicker/index.js
```diff
@@ -2,17 +2,32 @@ import PropTypes from 'prop-types';
...
 import { ContentSearch } from '../ContentSearch';
 import SortableList from './SortableList';
 
 const NAMESPACE = '10up-content-picker';
 
+/**
+ * Unfortunately, we had to use !important because on PickedItem we couldn't @emotion/styled css
+ * as it was breaking sortability from react-sortable-hoc
+ */
+const StyleWrapper = styled('div')`
```
johnwatkins0, line 16: Moved this outside the component.

components/ContentPicker/index.js
```diff
@@ -54,64 +69,49 @@ const ContentPicker = ({
...
-			border: none !important;
-			cursor: default;
-
-			&:hover {
-				background: transparent;
-			}
+	// Run onPickChange callback when content changes.
+	useEffect(() => {
+		onPickChange(content);
+	}, [content, onPickChange]);
```
johnwatkins0, line 75: Moved all instances of `onPickChange` into this effect hook so that it doesn't have to be called in so many places. With this hook, it will be called whenever the content changes.

components/ContentPicker/index.js
```diff
@@ -54,64 +69,49 @@ const ContentPicker = ({
...
+		setContent((previousContent) => [
+			{
+				id: item.id,
+				type: item.subtype,
+			},
+			...previousContent,
+		]);
+	};
+
+	const onDeleteItem = (deletedItem) => {
```
johnwatkins0, line 87: Removed the `sortIndex` argument because it's not needed to filter the deleted item out of the content. All content types/modes (posts, terms, users) will have an `id` to use for filtering.

components/ContentPicker/index.js
```diff
@@ -54,64 +69,49 @@ const ContentPicker = ({
...
+		setContent((previousContent) => previousContent.filter(({ id }) => id !== deletedItem.id));
+	};
+
+	const excludeItems = useMemo(() => {
```
johnwatkins0, line 91: Wrapped this in `useMemo` so that we're not re-creating this array on every render.

johnwatkins0:
Changed base branch to `develop`

tlovett1 (approved):

PR #44, opened 2021-08-17
Reset UI state on error

mitogh:
### Description of the Change

In case an error occurs on the API rest the UI into an empty state with no results.

Ideally, this should display an error or use a notice to display this error, but just make sure the UI is usable in case an error happens when fetching the data from the API.

**Previous Behavior**

**New Behavior**

### Alternate Designs

Another solution would be to use a Notice to display the error using the Gutenberg Notice store, but for now, this would allow the UI to keep on working.

### Benefits

Provide a usable UI after an error.

Video:

- https://share.vidyard.com/watch/a48t1f7tHKevRZRgCteuGq?

### Possible Drawbacks

Users might not be aware errors are happening unless it opens the console.

### Verification Process


components/ContentSearch/index.js
```diff
@@ -88,6 +88,10 @@ const ContentSearch = ({ onSelectItem, placeholder, label, contentTypes, mode, e
 
 				setSearchResults(newResults);
 				setIsLoading(false);
+			}).catch( (error) => {
```
johnwatkins0: @mitogh What do you think about giving the component a new piece of state, like:

```JSX
const [error, setError] = useState();
```

And then using the error from `catch` to display a message below the text area. The message could disappear once the user starts typing again. 

Using a notice, as you mentioned, might be a good idea, but this component is meant to be useful outside the editor context (e.g., on other admin screens), so notices might not always be available.

johnwatkins0: *Optional*  A little outside the scope of the PR, but what do you think about making this callback `async` and handling this error with `try`/`await`/`catch`/ instead of `then`/`catch`?

johnwatkins0 (commented):
Good idea. Works as is, but I left a couple of suggestions to at least consider.

components/ContentSearch/index.js
mitogh: I think having a state to manage the errors is the way to go I can polish this a bit more to make it happen and use await as well.

In terms of UI I can add the error message at the top of the component as default behavior but allowing other behavior if provided via a prop.

Thanks.

tlovett1: Agreed an error message would be good. Also, this will probably need to be reconfigured based on some other changes in trunk now.

mitogh: Thanks let me take over the conflicts and UI errors.

mitogh: Here's a vide of how it looks like when being used: 

- https://share.vidyard.com/watch/a48t1f7tHKevRZRgCteuGq?

tlovett1: Is this ready for re-review?

mitogh: Yes this should be ready for review, thanks.

tlovett1:
Hey @mitogh this is a really great idea. However, I'm having a lot of issues. I'm getting errors on every search. I think it's showing an error every time a request is aborted which is on every search.

fabiankaegy:
Closing this due to inactivity for now. We can always reopen when we run into the issue again.

PR #54, opened 2021-10-25
fix/23: duplicate key error

### LINKED ISSUE #23: Same items load with the same key and breaks sorting.
**Describe the bug**
The ContentPicker allows you to select the same post twice. But if you do that and reload, the items load with the same key and you get the following error in the console. "Warning: Encountered two children with the same key, `item-32`".

**Steps to Reproduce**

- Add a block using the Content Picker that has `isOrderable` set to true.
- Search for and select the same post twice.
- Save.
- Reload.
- You will see the error in the console. and you will not be able to click and drag to sort the items.

**Expected behavior**
User should not be able to select the same post twice.

Sidsector9:
Closes #23 

### Description of the Change
As per the conversation in the issue, instead of disallowing multiple items, error thrown for duplicate keys has been fixed.

The fix implements generation of unique key outside the loop because sorting/re-ordering items will not sort the keys for duplicate posts if they're adjacent to each other.

### Verification Process
Follow the steps in the issue and additionally set the following 2 props:
- `maxContentItems={ 6 }`
- `uniqueContentItems={ false }`

### Screencast

https://user-images.githubusercontent.com/17757960/138648303-5ebe6bc2-825b-4e62-a7e1-0a7ccfc312ae.mov

### Checklist:

### Changelog Entry
Fix: duplicate keys error for duplicate content items.

### DIFF
--- a/components/ContentPicker/PickedItem.js
+++ b/components/ContentPicker/PickedItem.js
@@ -54,14 +54,31 @@ const PickedItem = ({ item, isOrderable, handleItemDelete, mode }) => {
 		(select) => {
 			const { getEntityRecord, hasFinishedResolution } = select('core');
 
-			const getEntityRecordParameters = [type, item.type, item.id];
+			let itemId = item.id;
+
+			if (item?.duplicateOf === undefined) {
+				itemId = item.id;
+			} else if (item.duplicateOf > 0) {
+				itemId = item.duplicateOf;
+			}
+
+			const getEntityRecordParameters = [type, item.type, itemId];
+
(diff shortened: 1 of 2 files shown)


tlovett1:
Hey @Sidsector9 I think this PR might be a bit overly complicated. I created an alternative solution here - https://github.com/10up/block-components/pull/61. Let me know what you think.

Sidsector9:
@tlovett1 I approved #61 as it has a better approach 🙌 . Closing this one in favor of 61.

PR #56, opened 2021-11-03
Add `registerBlockExtention` API

fabiankaegy:
### Description of the Change

Following up on an internal blog post about "Pitfalls of Custom Block Styles" here: https://internal.10up.com/blog/2021/11/03/the-pitfall-of-custom-block-styles-in-wordpress/ this is a wrapper API to make adding these custom options to blocks less cumbersome. Instead of having to write 4 manual filters and a lot of boilerplate code this wrapper abstracts a lot of this away behind a more developer-friendly API. 

 The `registerBlockExtention` api is a syntactical sugar wrapper to make it easier to add custom settings that produce classnames to any blocks. The reason for this api is a limitation in custom block styles that mean that in many cases it is better to add new settings to blocks than it is to use block styles. This API is here to make that easier

#### Usage
```js
import { registerBlockExtention } from '@10up/block-components';

/**
 * BlockEdit
 *
 * a react component that will get mounted in the Editor when the block is
 * selected. It is reccomended to use Slots like `BlockControls` or `InspectorControls`
 * in here to put settings into the blocks toolbar or sidebar.
 *
 * @param {object} props block props
 * @returns {JSX}
 */
function BlockEdit(props) {...}

/**
 * generateClassNames
 *
 * a function to generate the new className string that should get added to
 * the wrapping element of the block.
 *
 * @param {object} attributes block attributes
 * @returns {string}
 */
function generateClassNames(attributes) {...}

registerBlockExtention(
	'core/group',
	{
		extentionName: 'background-patterns',
		attributes: {
			hasBackgroundPattern: {
				type: 'boolean',
				default: false,
			},
			backgroundPatternShape: {
				type: 'string',
				default: 'dots',
			},
			backgroundPatternColor: {
				type: 'string',
				default: 'green'
			}
		},
		classNameGenerator: generateClassNames,
		Edit: BlockEdit,
	}
);
```

#### Options
| Name                       | Type       | Description                                       |
|----------------------------|------------|---------------------------------------------------|
| blockName                  | `string`   | Name of the block the options should get added to |
| options.extentionName      | `string`   | Unique Identifier of the option added    |
| options.attributes         | `object`   | Block Attributes that should get added to the block |
| options.classNameGenerator | `funciton` | Funciton that gets passed the attributes of the block to generate a class name string |
| options.Edit               | `funciton` | BlockEdit component like in `registerBlockType` only without the actual block. So onyl using slots like the `InspectorControls` is advised. |
### Alternate Designs

/ 
### Checklist:


joesnellpdx:
@fabiankaegy I do think the readme needs more guidance. In my mind, I'd assume the blog post doesn't exist.

tlovett1 (approved):

PR #60, opened 2021-11-08
SearchItem - Add tooltip on hover

### LINKED ISSUE #17: Content Picker: at narrow widths, too little of the post title is visible

When using the content picker within narrow containers such as the block editor sidebar, it can be difficult to find the post you're looking for because the titles are cut off after the first couple of words. Screenshot:

Screen Shot 2021-03-16 at 4 56 56 PM

I know the CSS can be overridden within my project to make more of the title show (it's currently hidden with CSS), but by default within the component I suggest we surface the title either as a browser-native title attribute (see below) or with a similar Gutenberg component.

Screen Shot 2021-03-16 at 4 59 42 PM

Additionally, we might revisit the "post" badge shown to the right of the button, particularly when we are searching only one post type, in which scenario it adds no meaningful information.

xavortm:
Fixes #17 

### Description of the Change
Added Gutenberg-native tooltip on hover. Changed overflow-y to auto instead of scroll to hide the scrollbar when there is no need to scroll (few items to show)

Also "post" is not highlighted on hover now to increase contrast ratio

### Possible Drawbacks
Still, the titles are not fully visible. A potential solution would be to have a prop that says "Show full titles".

### Checklist:

### DIFF
--- a/components/ContentSearch/SearchItem.js
+++ b/components/ContentSearch/SearchItem.js
@@ -1,7 +1,24 @@
 import PropTypes from 'prop-types';
+import styled from '@emotion/styled';
 import { safeDecodeURI, filterURLForDisplay } from '@wordpress/url';
 import { decodeEntities } from '@wordpress/html-entities';
-import { Button, TextHighlight } from '@wordpress/components';
+import { Button, TextHighlight, Tooltip } from '@wordpress/components';
+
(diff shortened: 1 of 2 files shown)


tlovett1:
Works well for me. Nicely done.

tlovett1 (approved):

PR #61, opened 2021-11-08
Fix duplicate keys issue in content picker. Addresses #23

### LINKED ISSUE #23: Same items load with the same key and breaks sorting.

tlovett1:
Addresses #23. Creates a unique key for each picked post using the `uuid` package. This is an alternative to #54


Sidsector9 (approved):
@tlovett1 this is a better approach compared to #54. Tested on local and working as expected.

PR #66, opened 2021-11-19
Add: `Link` Component

### LINKED ISSUE #62: Add: `Link` Component

A common pattern that we need to solve for in our blocks is the ability to have one or multiple CTA/Link areas inside of a Custom Block. These CTAs/Links may be optional but when an editor chooses to use the Link they need to provide both the Link Text **and** also the URL. 

Create a new Component called `Link` that allows developers to pass a `text`, `url`, `onTextChange`, `onUrlSelect` and other props that may be passed through to the underlying `RichText` component like a `className etc.`

```js
const { linkText, linkUrl } = attributes;

<Link 
    text={linkText}
    onTextChange={ value => setAttributes( { linkText: value } ) }
    placeholder={ "Insert Link here..." }
    url={linkUrl}
    onLinkSelect={ value => setAttributes( { linkUrl: value } ) }
    className={`custom-class-name`}
/>
```

The Component should be aware of whether the current block is selected and not show when no text and no URL are provided and the block is not currently selected. When either of the values is shown but the other one isn't there should be an inline warning indication letting the user know that they are not done entering data into the component. 

When a user interacts with the link in the editor the `LinkControl` component should open up and allow editors to choose what they want to link to. 

**Designs**

/
 
**Describe alternatives you've considered**

/

**Additional context**

/

xavortm:
### Description of the Change

Add a `Link`component that makes it easy for developers to add link controls that work similarly to the core button block but within their own custom blocks. 

```js
 import { Link } from '@10up/block-components';
 const BlockEdit = (props) => {
     const { linkText, linkUrl, opensInNewTab } = attributes;
     const blockProps = useBlockProps();
      const handleTextChange = value => setAttributes({text: value});
     const handleLinkChange = value => setAttributes({
         linkUrl: value?.url,
         opensInNewTab: value?.opensInNewTab,
         linkText: value?.title ?? linkText
     });
     return (
         <div {...blockProps}>
             <Link 
                 value={ linkText }
                 url={ linkUrl }
                 opensInNewTab={ opensInNewTab }
                 onTextChange={ handleTextChange }
                 onLinkChange={ handleLinkChange }
                 className='example-classname'
                 placeholder='Enter Link Text here...'
             />
         </div>
     )
 }
 ```

Markup wise this will appear as just an anchor tag just as you would use it on the frontend of the site. However, when you as the user interact with the link it opens the Link control component in a popup to allow the user to change the link and other properties like opening in a new tab etc. 

### Rationale:

It's a common use case for custom blocks to include Call to Action links or links in general. The issue is creating a good consistent UI for it that gets out of the way but also shows what's needed. This component does just that. You can use as many `Link` components as you'd like in your block. 

Closes: #66 

Props: @xavortm & @fabiankaegy

### CODE REVIEW

**Comment by xavortm** [comment]
CleanShot 2021-11-22 at 14 14 06

With latest change the input shows the popup, but doesn't keep the focus on the link. Clicking opens the popup and ESC closes it.

**Comment by fabiankaegy** [comment]
Hey @xavortm 

I finally was able to take a look at this this morning and think this is a great start :) There are a few things that I think we can improve but overall this is already very solid. 

The main two points that I am thinking about are:

**1. How can we make it easier to edit the text of the link?** 
Right now the issue I'm seeing here is that as soon as you click on the link text you get the popover to appear with the link controls. Which is great on one hand but since the search bar in the LinkControl pulls focus makes it almost impossible to actually change the text of the rich text element. 

I'm not sure I know the best way around this jet but maybe we can tie it together with point 2.

**2. It would be great to add the inline validation like the original proposal from Helen in #48 had**
One of the biggest pain points of URL pickers in the admin is editors not filling in both the link text and destination. To make this easier the original component added inline validation using icons and tooltips to give the user the info that they are missing something. 

I wonder whether we can use that to our advantage for the first point I raised. When we always show a link icon in the selected state of the block we can make that the click target to trigger the popover. 

Thanks again for working on this and I am really excited for having this component in out toolbox :)

File: components/Link/index.js [hunk 1]
```diff
@@ -0,0 +1,215 @@
...
+const Link = ({
+	label,
+	type,
+	opensInNewTab,
+	url,
+	onLinkChange,
+	onTextChange,
+	kind,
+	placeholder,
+	isSelected,
```
fabiankaegy: Instead of relying on the developer to pass the `isSelected` prop here we can use the `useBlockEditContext` hook to get access to the current block from within our component.

```js
import { useBlockEditContext } from '@wordpress/block-editor';

const { isSelected } = useBlockEditContext();
```

components/Link/index.js
```diff
@@ -0,0 +1,215 @@
...
+	}
+}
+
+/**
+ * Link component that can be used inside other Gutenberg blocks for setting up URLs.
+ *
+ * The link should not be visible if the block is not focused. This will mainain nicer
+ * visuals in the block editor as a whole.
+ *
+ * @param props {object}
```
fabiankaegy: Can we please add the types for the object properties as well?

components/Link/index.js
xavortm: Thanks! This was updated

components/Link/index.js
xavortm: Aye, added for all + propTypes for required fields as the component wouldn't work too well if some props are not passed. Meaning, it would be easy to forget to pass `opensInNewTab` for example.

xavortm:
Thanks for the feedback!

I think I covered most of the issues. It is not done yet, but I am not 100% sure what is the best way to proceed. Meaning - should we have docs showing how the Link is updated from the callback functions and what props are provided. See how 
the `onLinkChange` gives an object for example which also provides the new tab value.

## Here are some video examples of how it works now


With URl provided in the block as a prop:
CleanShot 2021-12-06 at 13 14 25

Without URL provided, warning remains even if text is entered:
CleanShot 2021-12-06 at 13 15 46

I also added PropTypes to the component and also the JSDocs comments. I have called the Link like this:
`<Link onTextChange={(text) => {setValue(text)}} opensInNewTab={true} onLinkChange={(s)=> {setUrl(s.url);}} value={value} url={url.url} />`

Note that `onLinkChange` provides an object with this shape:
```jsx
const link = {
	url,
	opensInNewTab,
	title: value && navStripHTML(value), // don't allow HTML to display inside the <LinkControl>
};
```

components/Link/index.js
```diff
@@ -0,0 +1,242 @@
...
+import { Popover, Icon, Tooltip } from '@wordpress/components';
+import { isKeyboardEvent, ENTER } from '@wordpress/keycodes';
+import {
+	__experimentalLinkControl as LinkControl,
+	useBlockProps,
+	RichText,
+	useBlockEditContext,
+} from '@wordpress/block-editor';
+
+/**
```
fabiankaegy: I think this custom function can be replaced by adding `__unstablePastePlainText` to the `RichText` Component

components/Link/index.js
```diff
@@ -0,0 +1,242 @@
...
+				return { type: 'post', subtype: type };
+			}
+			return {};
+	}
+}
+
+const LinkOutput = styled.a`
+	--color--warning: #f00;
+
+	color: var(--wp--style--color--link, var(--global--color-primary));
```
fabiankaegy: where are we getting the fallback custom property `--global--color-primary` from? That one does not seem familiar to me.

components/Link/index.js
```diff
@@ -0,0 +1,242 @@
...
+
+	const link = {
+		url,
+		opensInNewTab,
+		title: value && navStripHTML(value), // don't allow HTML to display inside the <LinkControl>
+	};
+
+	const blockProps = useBlockProps({
+		ref,
+		className: classnames('wp-block-navigation-item', {
```
fabiankaegy: I'm curious about the choice of class name here.

`wp-block-navigation-item` does not seem like the right choice to me. I would like to adhere closely to how core names their components and therefore think something like `components-link` would be better suited to this element. We can namespace it to be `tenup-block-components-link`

components/Link/index.js
```diff
@@ -0,0 +1,242 @@
...
+	 */
+	useEffect(() => {
+		setIsValid(!!url && !!value);
+	}, [url, value]);
+
+	return (
+		// eslint-disable-next-line react/jsx-props-no-spreading
+		<span {...blockProps}>
+			{/* eslint-disable jsx-a11y/anchor-is-valid */}
+			<LinkOutput>
```
fabiankaegy: I wonder whether we could by without this wrapper. Essentially I'm wondering if we can get the markup to closer resemble the frontend markup. 

My main issue is that we are nesting a block level element inside of the inline anchor tag. 

We can ressolve this by either setting the `RichText` to `tagName="a"` and then add the custom styling for in in css rather than a styled component. 

I think it would also be fine to switch the `RichText` to `tagName="span"` so it at least is an inline element inside the inline link

components/Link/index.js
```diff
@@ -0,0 +1,242 @@
...
+					placeholder={placeholder}
+					withoutInteractiveFormatting
+					allowedFormats={['core/bold', 'core/italic', 'core/strikethrough']}
+					onClick={() => {
+						setIsLinkOpen(true);
+					}}
+				/>
+
+				{!isValid && (
+					<Tooltip text="URL or Text has not been set">
```
fabiankaegy: This should be translatable

components/Link/index.js
```diff
@@ -0,0 +1,242 @@
...
+			</LinkOutput>
+		</span>
+	);
+};
+
+Link.defaultProps = {
+	type: '',
+	kind: '',
+
+	placeholder: 'Link text ...',
```
fabiankaegy: this should be translatable

fabiankaegy (changes requested):
Thank you for all the work you have already put into this! 

There are a few things that I have found while tearing the component that I think we need to resolve before merging this. 
1. The component is not jet exported from the [index.js](https://github.com/10up/block-components/blob/develop/index.js) file of the repository.
2. The Popover does not close when the user clicks outside of the Link Component.
CleanShot 2021-12-14 at 10 47 47
I think the preferred solution for this would be to close the popover when the user clicks outside of the modal except for when they click on the link.
3. We should add some documentation with code examples to the readme
4. Right now you can only change the `openInNewTab` value through code. I would like to be able to also make that user changeable. The `LinkControl` allows you to change this value and store the value in the blocks attributes. 

Outside _believe_ that we should be able to replace the custom `navStripHTML` function by setting the `__unstablePastePlainText` prop of the `RichText` Component.

I've added a few more inline comments with questions related to the code :)

xavortm:
Thank you for the review @fabiankaegy , I will go through all of the points and try to polish anything I find in addition in the next few days. There is some leftover code from what I have reused from other parts of the editor:3

xavortm:
At the moment, it's possible to update the `opensInNewTab` like this: https://gist.github.com/xavortm/747b00e116280096c2ca2700c9d3cd2b 

I don't know how to make the closing of the popup :( I was debugging and testing for about 2 hours, but no success. I can't find how exactly the regular link in a paragraph works. It closes the popup once you move out of the editing <a> tag, but keeps focus when you click on the popup (as we want it). 

In our case, we use the isSelected hook, but the problem is that this looks for the block in use, not the component. So, as long as the block you add the Link to is selected, the popup will show up...

Do you have ideas on what can be tried here? Aside from that, I think I covered most other smaller points without the documentation. I would work on that after the component seems to be working well.

fabiankaegy:
I've merged my branch into this one and will take it from here :)

fabiankaegy:
@xavortm I have made some changes to how the popover works. It now resembles the behavior of the core link rich text format much more and just stays open as long as you are interacting with the linkable element. 

I would love to get your review of this PR before we pass it over to TL for the final review :) Let me know if you won't be able to get to it in the next week or so :)

fabiankaegy (commented):
@tlovett1 I've updated the description of the issue to better reflect the rationale for the component. Please let me know if you have any questions.

tlovett1 (approved):

PR #71, opened 2021-12-23
ContentPicker - Add load more button

### LINKED ISSUE #51: Content Picker: Paginate results or use Load More button; Avoid inifinte scroll.
The content picker results modal should follow WordPress' existing patterns such as pagination or a load more button. 

Pagination would show 10 or a configurable number of results upon search then paginate accordingly. Or search results should follow the new Media Library design pattern that displays a `Load More` button for the visitor to initiate the next set of results. In both cases, infinite scroll should be avoided due to accessibility limitations summarized in this [post](https://core.trac.wordpress.org/ticket/40330)

xavortm:
### Description of the Change

Add load more button to results of the ContentPicker. Why is described in #51 I am not super happy about the end code though... I am using this for React practice, so please if there are any issues with the code overall, let me know and I will try to fix them after the CR ^^ Biggest change I see is that the query is moved into a useEffect hook and I am running it on state change instead of from function call.

No "load more", meaning, everything has loaded: 
Screenshot 2021-12-23 at 15 50 56

Has "load more", meaning, more posts to load: 
Screenshot 2021-12-23 at 15 52 17

PR has some code formatting that came from prettier. I have also reduced the search results from 50 to 20.

### Alternate Designs

Paginated results, so we have a set of pages to click. I wasn't sure how that would be implemented, so I went for a load more button.

### Checklist:


tlovett1:
Hey @xavortm really nice work! I have some feedback:
1. Load More button styling does not look right. I would just apply the classes `components-button is-secondary`. It should also be centered.
2. When you type in a new search string, it shows the loading spinner and the loading... where the button is. It should only show the spinner in that state.
3. It appears load more isn't actually checking the API response headers to determine if there are more pages. The way you are currently doing it won't work if there are exactly 20 results - it would show the load more button.

xavortm:
Thank you! For some reason the button props didn't work out for me so Indeed I can just use classes 👍  I will see how the API check works and get a patch there hopefully in the coming 1-2 weeks

xavortm:
Updated the position and styling of the load more as well as the box in general to take the available space: [Screenshot](https://cln.sh/htR9ckwU1gQrEj3ixnyL)

I have also changed:

- Added min-width to the load more button so that when it changes to "Loading" its width doesn't change.
- Added min-height to the "Spinner" component to match the default min-height of the <li> results. This will also reduce the jumping when loading the first result.
- Added wrapper around the spinner to center it in the middle. 

And the one that took me the most time was the third point of the feedback to show load more based on how many posts are in the results. For this, I had to change a bit of request and removed the parsing in the options to expose the headers information.

I also updated the caching mechanics a bit to support the paginated content, because, without this, we wouldn't have "load more" that is synced with the amount of content that can be loaded and is cached.

Manual test cases with 20+ posts that will return when searching "1":

- Search query "1"
- Hit "Load more"
- Delete search query
- Search "1"
- Expected: load first page of cached results
- Hit "load more"
- Expected: load first + second page from fetch
- Hit "Load more"
- Expected: load first + second + third page results from fetch

tlovett1:
FYI, I fixed a bug where the picker component was non-stop re-rendering.

components/ContentPicker/index.js
```diff
@@ -87,7 +87,7 @@ const ContentPicker = ({
 	// Run onPickChange callback when content changes.
 	useEffect(() => {
 		onPickChange(content);
-	}, [content, onPickChange]);
+	}, [content]);
```
fabiankaegy, line 94: @tlovett1 This will likley cause other issues because of the omitted dependency. The correct approach would be making the `onPickChange` a callback using the `useCallback` hook: https://reactjs.org/docs/hooks-reference.html#usecallback

tlovett1: I will defer to React guru @nicholasio on this one. I think since it's passed via prop from an unknown place, this is fine.

nicholasio: I don't think `useCallback` here is the best option as it's essentially going to expect that users of this library are using it and if they aren't there will be a non-stop re-rendering issue.

The easiest way to solve this is to remove the dep and disable the eslint rule and it should be mostly fine.

However, I took a quick look at the code and a better solution to this issue is to stop using `useEffect` as a way to watch for content changes and instead use the `setContent` callback to trigger `onPickChange`.

```javascript
setContent((previousContent) => { 
  const newContent = [
	  {
		  id: item.id,
		  uuid: uuidv4(),
		  type: 'subtype' in item ? item.subtype : item.type,
	  },
	  ...previousContent,
  ];
 onPickChange(newContent)
 return newContent;
});
```
This would need to be done on every setContent call

tlovett1: Got it. I'd like to leave it as it is now in the interest of changing less :)

fabiankaegy:
@tlovett1 @xavortm I resolved the merge conflicts. 

However I am getting an odd behavior from the load more feature even before merging `develop` into it.

When I search for something and have clicked on the load more button The old results never disappear. I just get more added to the bottom of the list. 

https://user-images.githubusercontent.com/20684594/154213773-1b081d65-f688-4f6c-96af-631fac7686a9.mp4

tlovett1:
I just pushed a major overhaul of how we handle search state. This fixes a bunch of problems I was having mostly around race conditions. @nicholasio @fabiankaegy would love for you to take a look.

fabiankaegy (approved):
I tested this locally and it works like a charm 👍

PR #73, opened 2022-01-12
Fix: Link Component Popover state

fabiankaegy:
This PR enhances the work in #66 by adding a `useOnClickOutside` hook and cleaning up how the component manages it's popover state

### DIFF
--- a/components/Link/index.js
+++ b/components/Link/index.js
@@ -1,7 +1,6 @@
 /**
  * External dependencies
  */
-import classnames from 'classnames';
 import styled from '@emotion/styled';
 import PropTypes from 'prop-types';
 
@@ -9,15 +8,16 @@ import PropTypes from 'prop-types';
  * WordPress dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { useState, useEffect, useRef, Fragment } from '@wordpress/element';
+import { useState, useEffect, useRef } from '@wordpress/element';
 import { Popover, Icon, Tooltip } from '@wordpress/components';
-import { isKeyboardEvent, ENTER, ESCAPE } from '@wordpress/keycodes';
 import {
 	__experimentalLinkControl as LinkControl,
-	useBlockProps,
 	RichText,
 	useBlockEditContext,
 } from '@wordpress/block-editor';
+
+import classNames from 'classnames';
+import { useOnClickOutside } from '../../hooks/use-on-click-outside';
 
 /**
  * Given the Link block's type attribute, return the query params to give to
@@ -63,7 +63,7 @@ const LinkOutput = styled(RichText)`
 	gap: 0.5em;
 	text-decoration: underline;
 
-	/* This  holds the text URL input */
+	/* This holds the text URL input */
 	& > div {
 		text-decoration: underline;
 	}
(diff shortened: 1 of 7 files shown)


components/Link/index.js
```diff
@@ -155,21 +150,17 @@ const Link = ({
...
-				{...blockProps}
 				tagName="a"
 				identifier="label"
-				className="tenup-block-components-link__label"
+				className={classnames('tenup-block-components-link__label', className)}
 				value={value}
 				onChange={onTextChange}
 				aria-label={__('Link text', '10up-block-components')}
 				placeholder={placeholder}
-				withoutInteractiveFormatting
```
fabiankaegy, line 167: Since we are already setting the `allowedFormats` this had no effect. 

The `withoutInteractiveFormatting` removes any rich text formats that add interactivity to the string.

components/Link/index.js
```diff
@@ -155,21 +150,17 @@ const Link = ({
...
 				__unstablePastePlainText
 				allowedFormats={['core/bold', 'core/italic', 'core/strikethrough']}
-				onClick={() => {
```
fabiankaegy, line 170: This is now replaced by `onFocusCapture`

components/Link/index.js
```diff
@@ -181,19 +172,26 @@ const Link = ({
 			{isLinkOpen && (
 				<Popover
 					position="top center"
-					onClose={() => setIsLinkOpen(false)}
```
fabiankaegy, line 184: This was redundant since the `onClose` callback is only invoked when the popover gets unmounted

components/Link/index.js
```diff
@@ -202,21 +200,23 @@ const Link = ({
...
+	url: undefined,
+	className: undefined,
 	type: '',
 	kind: '',
-
 	placeholder: __('Link text ...', '10up-block-components'),
 };
 
 Link.propTypes = {
-	value: PropTypes.string.isRequired,
```
fabiankaegy, line 212: The initial value of an attribute is `undefined` if no default value is set. Therefore the `isRequired` caused a console error.

components/Link/index.js
```diff
@@ -109,38 +113,29 @@ const Link = ({
...
+
+	const popoverRef = useOnClickOutside(closePopover);
 
 	const link = {
 		url,
 		opensInNewTab,
 		title: value, // don't allow HTML to display inside the <LinkControl>
 	};
 
-	const blockProps = useBlockProps({
```
fabiankaegy, line 124: The useBlockProps hook is really only meant for blocks itself and the props should only be used on the outer most wrapper element. Anything else will lead to undesired behaviors

components/Link/index.js
```diff
@@ -9,16 +9,19 @@ import PropTypes from 'prop-types';
  * WordPress dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { useState, useEffect, useRef, Fragment } from '@wordpress/element';
+import { useState, useEffect, useRef } from '@wordpress/element';
 import { Popover, Icon, Tooltip } from '@wordpress/components';
-import { isKeyboardEvent, ENTER, ESCAPE } from '@wordpress/keycodes';
```
fabiankaegy, line 14: These values were not used

PR #78, opened 2022-01-17
feat/75: Add Optional Component

### LINKED ISSUE #75: Add: `Optional` Component
One pattern that almost every block should use is showing placeholders of optional values when a block is selected but hiding them when the block is not selected. 

Currently the only way to get that behavior is by checking that the value itself is empty and checking that the block is not selected:

```js
const BlockEdit = (props) => {
    const { attributes, setAttributes, isSelected } = props;
    const { title } = attributes;

    const blockProps = useBlockProps();

    return (
        <div {...blockProps}>
            { !! title && isSelected && (
                <RichText tagName="h2" value={ title } onChange={value => setAttributes({ title: value }) } />
            ) }
        </div>
    )
}
```

Instead of always having to do this how about we add a component called `Optional` that takes care of this logic for us. The component itself can use the [`useBlockEditorContext`](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/README.md#useblockeditcontext) to get the selection state and we can pass it only the value to check:

```js
const BlockEdit = (props) => {
    const { title } = attributes;

    const blockProps = useBlockProps();

    return (
        <div {...blockProps}>
            <Optional value={ title }>
                <RichText tagName="h2" value={ title } onChange={value => setAttributes({ title: value }) } />
           </>
        </div>
    )
}
```

The main benefit here is legibility. It makes it more clear what you are trying to achieve and therefore makes it simpler to compose.

Sidsector9:
### Description of the Change
Adds the `<Optional />` Component which renders its children only when it is selected.

Closes #75 

### Screencast

https://user-images.githubusercontent.com/17757960/149763950-d2402d80-b3e1-4cf0-b158-93a6c0934013.mov

### Checklist:

### Changelog Entry
Added - Optional Component

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,7 +279,7 @@ The `<RichText>` node will only render when BlockEdit is selected.
 #### Props
 | Name       | Type              | Default  |  Description                                                   |
 | ---------- | ----------------- | -------- | -------------------------------------------------------------- |
-| `value` | `string`    | `''`   | The value that will be consumed by the children |
+| `value` | `string`    | `''`   | The value that will be consumed by the children. If the value is falsey the component will only be rendered if the block is selected. |
 
 ## registerBlockExtention
 The `registerBlockExtention` API is a wrapper to make it easier to add custom settings which produce classnames to any blocks. There are a few problems with using block styles for customisations. For one an editor cannot combine block styles. So you very quickly land in a sittuation where you need to add many block styles just to give an editor the ability to choose exactly the combination of options they want. That leads to a bad user experience though as the previews take up a ton of space and also make the editor slower due to the overhead of the iframes it creates. So in many cases it is nicer to extend a bock with custom settings to achive the same goal. The process of registering your own attributes, modifying the blocks edit function, adding the new classname to the editor listing and also adding it to the frontend is rather cumbersome though. That is where this API comes in. It is a wrapper for the underlying filters that improves the editorial experience and reduces the amount of code that needs to get maintained in order to extend blocks.
--- a/README.md
+++ b/README.md
@@ -253,6 +253,28 @@ const MyComponent = ({clientId}) => {
 | `slidesPerPage` | `integer` | `1` | Number of slides to show per page |
 | `parentBlockId` | `string` | `''` | Client ID of parent block. This is required.  |
 
+## Optional
+A component that takes care of the logic of rendering nodes only when it is selected.
+
+### Usage
+```js
+const BlockEdit = (props) => {
+    const { attributes, setAttributes, isSelected } = props;
+    const { title } = attributes;
+
+    const blockProps = useBlockProps();
+
+    return (
+        <div {...blockProps}>
+            <Optional value={ title }>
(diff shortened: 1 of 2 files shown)


components/Optional/index.js
```diff
@@ -0,0 +1,16 @@
+import PropTypes from 'prop-types';
+import { useBlockEditContext } from '@wordpress/block-editor';
+
+export const Optional = ({ value, children }) => {
+	const { isSelected } = useBlockEditContext();
+	return !!value && isSelected && <>{children}</>;
```
fabiankaegy:
```suggestion
	return !!value && isSelected && children;
```

I think we can omit the fragment here :)

README.md
```diff
@@ -253,6 +253,28 @@ const MyComponent = ({clientId}) => {
 | `slidesPerPage` | `integer` | `1` | Number of slides to show per page |
 | `parentBlockId` | `string` | `''` | Client ID of parent block. This is required.  |
 
+## Optional
+A component that takes care of the logic of rendering nodes only when it is selected.
```
fabiankaegy, line 257: I'd love for this description to explain what the value prop is for and how it should be used :)

fabiankaegy (changes requested):
This is awesome :) Thanks for working on this!!! :) I left two comments that I think can be improved but once they are fixed this is ready to be merged  👍

README.md
```diff
@@ -253,6 +253,33 @@ const MyComponent = ({clientId}) => {
...
+    )
+}
+```
+
+The `<RichText>` node will only render when BlockEdit is selected.
+
+#### Props
+| Name       | Type              | Default  |  Description                                                   |
+| ---------- | ----------------- | -------- | -------------------------------------------------------------- |
+| `value` | `string`    | `''`   | The value that will be consumed by the children |
```
fabiankaegy:
```suggestion
| `value` | `string`    | `''`   | The value that will be consumed by the children. If the value is falsey the component will only be rendered if the block is selected. |
```

components/Optional/index.js
```diff
@@ -0,0 +1,16 @@
+import PropTypes from 'prop-types';
+import { useBlockEditContext } from '@wordpress/block-editor';
+
+export const Optional = ({ value, children }) => {
+	const { isSelected } = useBlockEditContext();
+	return !!value && isSelected && children;
```
fabiankaegy: Sorry I missed this in the first round. This actually is not working as expected. 

The children should always get rendered if the block is selected. Only when the block is not selected and the value is falsey should the component be hidden. 

```suggestion
	return ( isSelected || !!value ) && children;
```

fabiankaegy (changes requested):
Sorry, missed that in the first run :)

components/Optional/index.js
Sidsector9: I built the component following the example in the issue which had this logic along the lines - `!! title && isSelected && children` 
Should we update the issue description?

fabiankaegy (approved):
Looks great  👍 

Tagging @tlovett1 to do a final review and merge  👍

tlovett1 (approved):

PR #80, opened 2022-01-17
Feature: Add ClipboardButton component

### LINKED ISSUE #72: Add: `ClipboardButton` Component
In WordPress 5.9 the `ClipboardButton` from the `@wordpress/components` package has been deprecated in favor of the `useCopyToClipboard` hook. The hook is great and allows for much more flexibility. But I do think that it still makes sense to have a more easy to use component that creates a button to copy something. 

## Proposed API

```js
import { ClipboardButton } from '@10up/block-components';

<ClipboardButton 
    text="Text that should be copied"
    disabled={ false }
    onSuccess={ () => {} }
    label={ ({ hasCopied }) => hasCopied ? 'Copied' : 'Copy' }
/>
```

When a user clicks the ClipboardButton label is a callback function that passes whether the text has been copied as the first parameter. This can be used to set the label depending on the state of the button. When the text has been copied the state should stay in the copied state for 3 seconds and then get reset to the original state.

Sidsector9:
### Description of the Change

Adds a button-based component to copy the text passed to it to the clipboard.

Closes #72 

### Changelog Entry
Added - ClipboardButton component


Sidsector9:
Hi @fabiankaegy I've added a default method for label prop [here](https://github.com/10up/block-components/pull/80/files#diff-b7334d7442de336128ade86597ffaf2a7ee5172e6a3583c8e0e4557be1a57e70R46).

I'm not sure how to add this to the README props table. Any suggestions?

fabiankaegy:
@Sidsector9 I would add something like this: 

The `label` prop takes a callback function that will get the current state passed as the first and only parameter. This state can be used to determine whether the value has just been copied or is ready to be copied.

What we could also do if you think that is a better DX is renaming the `label` to `labels` and have it accept an object containing to values: `{ copied: "Copied", copy: "Copy" }`

Sidsector9:
Agreed the 2nd approach seems better DX-wise. I'll push the changes, thanks @fabiankaegy

Sidsector9:
Changes pushed 👍

components/ClipboardButton/index.js
```diff
@@ -0,0 +1,55 @@
...
+	const [hasCopied, setHasCopied] = useState(false);
+	const copy = labels.copy ? labels.copy : __('Copy');
+	const copied = labels.copied ? labels.copied : __('Copied');
+
+	useEffect(() => {
+		if (hasCopied) {
+			setTimeout(() => {
+				setHasCopied(false);
+			}, 3000);
+		}
```
fabiankaegy, line 20: we should make sure that we clean up the timeouts on unmount:

https://stackoverflow.com/questions/53090432/react-hooks-right-way-to-clear-timeouts-and-intervals

fabiankaegy (approved):
This is looking great  👍

fabiankaegy:
@tlovett1 requesting your final approval here before we merge

tlovett1 (approved):

PR #90, opened 2022-02-17
add github action that auto deploys @next version of npm package

fabiankaegy:
### Description of the Change

In order to be able to test newer versions of the block components having the `@next` version auto publish the current `develop` branch would be a huge win. This aims to add a GitHub Action that does just that. Whenever something is merged into the `develop` branch this will increase the npm prerelease version and publish the version with the `next` tag.

In order for this to actually work a new `Secret` called `NPM_TOKEN` needs to be added to this repository to allow the action to deploy to the npm repository. 

### Checklist:

### Credits

Props @

### DIFF
--- a/.github/workflows/release-next-npm-version.yml
+++ b/.github/workflows/release-next-npm-version.yml
@@ -1,0 +1,18 @@
+name: Publish @next version to NPM
+on:
+  push:
+    branch: [ develop ]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


fabiankaegy:
In order for this to work the `NPM_TOKEN` secret needs to be added to this Repository. It needs to be an `Automation` token since that doesn't require 2FA.

CleanShot 2022-02-17 at 13 44 42@2x

tlovett1 (approved):

PR #93, opened 2022-02-18
Switch package to ship a transpiled version

### LINKED ISSUE #79: ContentPicker component will not build when using @wordpress/scripts
### Describe the bug

When trying to use the ContentPicker component in a project that only uses the `@wordpress/scripts` package, I get error below. Being able to use this component without the need for 10upscripts would help drive adoption.

```
ERROR in ./node_modules/@10up/block-components/api/registerBlockExtention.js 65:11
Module parse failed: Unexpected token (65:11)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
|                       // return early from the block mofification
|                       if (name !== blockName) {
>                               return <BlockEdit {...props} />;
|                       }
| 
 @ ./node_modules/@10up/block-components/index.js 8:0-70 8:0-70
 @ ./src/index.js 2:0-55 4:33-46

ERROR in ./node_modules/@10up/block-components/components/ContentPicker/index.js 119:2
Module parse failed: Unexpected token (119:2)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| 
|       return (
>               <div className={`${NAMESPACE}`}>
|                       {(!content.length || (content.length && content.length < maxContentItems)) && (
|                               <ContentSearch
 @ ./node_modules/@10up/block-components/index.js 1:0-59 1:0-59
 @ ./src/index.js 2:0-55 4:33-46

ERROR in ./node_modules/@10up/block-components/components/ContentSearch/index.js 145:2
Module parse failed: Unexpected token (145:2)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| 
|       return (
>               <NavigableMenu onNavigate={handleOnNavigate} orientation="vertical">
|                       <TextControl
|                               label={label}
 @ ./node_modules/@10up/block-components/index.js 2:0-59 2:0-59
 @ ./src/index.js 2:0-55 4:33-46

ERROR in ./node_modules/@10up/block-components/components/CustomBlockAppender/index.js 35:2
Module parse failed: Unexpected token (35:2)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| }) => {
|       return (
>               <Inserter
|                       isAppender
|                       rootClientId={rootClientId}
 @ ./node_modules/@10up/block-components/index.js 7:0-82 7:0-82
 @ ./src/index.js 2:0-55 4:33-46

ERROR in ./node_modules/@10up/block-components/components/IconPicker/icon-picker-toolbar-button.js 25:2
Module parse failed: Unexpected token (25:2)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| export const IconPickerToolbarButton = (props) => {
|       return (
>               <Dropdown
|                       className="component-icom-picker-toolbar-button"
|                       contentClassName="component-icom-picker-toolbar-button__content"
 @ ./node_modules/@10up/block-components/components/IconPicker/index.js 3:0-71 3:0-71
 @ ./node_modules/@10up/block-components/index.js 11:0-16:33 11:0-16:33 11:0-16:33 11:0-16:33 11:0-16:33
 @ ./src/index.js 2:0-55 4:33-46

ERROR in ./node_modules/@10up/block-components/components/IconPicker/icon-picker.js 96:2
Module parse failed: Unexpected token (96:2)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| 
|       return (
>               <BaseControl label={label} id={id} className="component-icon-picker" {...rest}>
|                       <SearchControl value={searchTerm} onChange={setSearchTerm} id={id} />
|                       {hasIcons ? (
 @ ./node_modules/@10up/block-components/components/IconPicker/index.js 2:0-43 2:0-43
 @ ./node_modules/@10up/block-components/index.js 11:0-16:33 11:0-16:33 11:0-16:33 11:0-16:33 11:0-16:33
 @ ./src/index.js 2:0-55 4:33-46

ERROR in ./node_modules/@10up/block-components/components/IconPicker/icon.js 20:9
Module parse failed: Unexpected token (20:9)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| 
|       if (!icon) {
>               return <Spinner />;
|       }
| 
 @ ./node_modules/@10up/block-components/components/IconPicker/index.js 1:0-30 1:0-30
 @ ./node_modules/@10up/block-components/index.js 11:0-16:33 11:0-16:33 11:0-16:33 11:0-16:33 11:0-16:33
 @ ./src/index.js 2:0-55 4:33-46

ERROR in ./node_modules/@10up/block-components/components/IconPicker/inline-icon-picker.js 25:2
Module parse failed: Unexpected token (25:2)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
|       const { value, ...rest } = props;
|       const IconButton = ({ onToggle }) => (
>               <Icon name={value.name} iconSet={value.iconSet} onClick={onToggle} {...rest} />
|       );
| 
 @ ./node_modules/@10up/block-components/components/IconPicker/index.js 4:0-56 4:0-56
 @ ./node_modules/@10up/block-components/index.js 11:0-16:33 11:0-16:33 11:0-16:33 11:0-16:33 11:0-16:33
 @ ./src/index.js 2:0-55 4:33-46

ERROR in ./node_modules/@10up/block-components/components/InnerBlockSlider/index.js 92:2
Module parse failed: Unexpected token (92:2)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| 
|       return (
>               <div className="inner-block-slider">
|                       <div className="slides-outer">
|                               <div className="slides" css={slidesCSS} ref={slides}>
 @ ./node_modules/@10up/block-components/index.js 3:0-65 3:0-65
 @ ./src/index.js 2:0-55 4:33-46

webpack 5.66.0 compiled with 9 errors in 205 ms
```

### Steps to Reproduce

1. Create a new package.json
2. Run `npm i -D @wordpress/scripts`
3. Run `npm install --save @10up/block-component`
4. Create `start` and `build` scripts as [per the docs](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-scripts/#setup)
5. Create a file at `src/index.js`
6. Import the ContentPicker component using the example on the repo.
7. Run `npm start` or `npm run build`

_No response_

### Environment information

- Device: Mac Mini M1
- OS: Big Sur 11.6
- Env: .zsh terminal 

### WordPress information

_No response_

### Code of Conduct

fabiankaegy:
### Description of the Change

Currently the `10up/block-components` package ships without any transpilation. This is causing issues when this project is being used outside of the context of `10up-toolkit` which has a [special edge case for this package hardcoded into its Webpack config](https://github.com/10up/10up-toolkit/blob/9f6f6a8c5088bb072320d24828520a16eb92d0e5/packages/toolkit/config/webpack/modules.js#L41).

Looking into it the issue appears to be that webpack doesn't expect the JSX syntax within what it expects to be vanilla JS. In order to fix this and allow the usage of this package in more use cases the best approach would be to actually transpile the code. 

We don't however want to actually include the code of the `@wordpress/*` packages because that is already loaded in the WordPress admin and including it here would cause unwanted bloat and possible inconsistencies. Also, we do still want our transpiled code to include the `import { x } from '@wordpress/y';` imports in order for them to be picked up from the Dependency Extraction Webpack Plugin to handle the dependency management for WordPress core. 

Closes #79 

### Verification Process

In order to test this, I forked this project and released a [transpiled version of the package on NPM](https://www.npmjs.com/package/fabiankaegy-block-components-fork). You can install it on your project like so: 

```zsh
npm install --save fabiankaegy-block-components-fork@next
```

### Checklist:

### Credits

Props @


fabiankaegy:
@nicholasio I'd love to get your eyes on this here. Especially I'm wondering how to best include the `@wordpress` packages as webpack externals. 

I have currently listed them as `peerDependencies` but that does not seem correct and also I'd love to not include all of them manually there. Ideally, the list of externals would get auto-generated out of all the `@wordpress/` packages used in this package. 

_(Note: There are some [edge cases which are not bundled](https://github.com/WordPress/gutenberg/blob/a4bf9b68b81f78b477b9476021f2ddffeddabbfd/packages/dependency-extraction-webpack-plugin/lib/util.js#L2-L6). )_

https://github.com/10up/block-components/blob/62a487be277132cd76361751acd6cd8c5bb229eb/package.json#L46-L61

---

Another thing I'd love to get your thoughts on is tree shaking. Currently one of the issues that I am seeing here, is that when I import one singular package from the `10up/block-components` all of the '@wordpress/` dependencies used in any of the components get added to the list of dependencies generated from the Dependency Extraction Webpack Plugin. This is because the entire file is being imported instead of just the file including the source code for the particular component. 

This can very quickly cause bloat because it means that a lot of additional packages will always need to be loaded if any of these components get used.

nicholasio:
We should definitely remove this package from the `exclude` in 10up-toolkit webpack config. I believe we can do so in a way that doesn't break existing projects, worst-case scenarios projects with older versions of the toolkit and newer versions of this package would transpile the code twice.

Regarding the deps, since we don't want to bundle them NOR have the consumers install them we should make them a dev dependency. If we make them peer deeps, we're mistakenly telling the consumers that those packages are indeed needed. Additionally, npm >=7 already resolves peer deps automatically.

With that said, we should not be installing any of the `@wordpress` packages. Once you move them to a dev dep you can do the following to tell `10up-toolkit` to skip bundling them with `--externals` 

```
10up-toolkit build --external=@wordpress/url,@wordpres/date,....
```

Unfortunately, this syntax doesn't support something like `@wordpress/*`

However, I believe you could use the dependency extraction plugin to automatically make the wordpress packages external.

The way 10up-toolkit is set up right now only supports running that webpack plugin in project mode but you could create a new webpack config and just include the plugin

```
const config = require('10up-toolkit/config/webpack.config.js');
// you might get a eslint error for unresolved dep but this is already available from 10up-toolkit, you can eslint disable this line.
const DependencyExtractionWebpackPlugin = require('@wordpress/dependency-extraction-webpack-plugin');

config.plugins.push(
  new DependencyExtractionWebpackPlugin({
    injectPolyfill: true,
 })
);

return config;
```

fabiankaegy:
@nicholasio thanks for adding your thoughts here! 

The issue with using the dependency extraction plug-in here in this package is that it also transpiles the import statements to point to the window object. 

Which would then cause them to no longer get picked up by the actual projects dependency extraction plug-in. 

I think there might be a way to override that behavior though. Will look into that :)

webpack.config.js
```diff
@@ -0,0 +1,15 @@
...
+module.exports = {
+	...config,
+	plugins: [
+		...config.plugins,
+		new DependencyExtractionWebpackPlugin({
+			requestToExternal: () => undefined,
+			requestToHandle: () => undefined,
+		}),
+	],
+};
```
fabiankaegy: @nicholasio Do you have any idea what I might be doing wrong here? this is not getting picked up 🤔

fabiankaegy: Never mind -.-

fabiankaegy:
@nicholasio I think I have solved the issue of the externals with this approach here:

https://github.com/10up/block-components/blob/feature/transpiled-delivery/webpack.config.js

Now my other question regards tree-shaking of the packages. Do you have any thoughts on how we improve that aspect of this?

nicholasio:
@fabiankaegy do you know if webpack is tree-shaking the code itself? When you include a single component from this package does webpack include the whole package code in the final bundle?

fabiankaegy:
@nicholasio Just tested this by trying to only import the `IsAdmin` component like so: 

```js
import { IsAdmin } from 'fabiankaegy-block-components-fork';

export default function BlockEdit() {
	return (<IsAdmin><p>test</p></IsAdmin>)
}
```

and it does not do any treeshaking.

You can test the current transpiled version like so:
```bash
npm install --save fabiankaegy-block-components-fork@next
```

tlovett1:
@fabiankaegy this one is over my head. Before merging, I'd like @nicholasio to confirm this works and doesn't break anything.

package.json
```diff
@@ -41,7 +43,11 @@
     "uuid": "^8.3.2"
   },
   "devDependencies": {
+    "@wordpress/dependency-extraction-webpack-plugin": "^3.3.1",
     "10up-toolkit": "^2.1.1",
     "cypress-localstorage-commands": "^1.6.1"
+  },
+  "10up-toolkit": {
+    "libraryName": "@10up/block-components"
```
nicholasio: This is not needed as you're not creating a UMD bundle. If you were creating a UMD bundle the library name would be used to expose lib via a global var e.g: `window.TenUpBlockComponents`.

UMD bundles are only created if you specify `umd:main` field in package.json.

webpack.config.js
```diff
@@ -0,0 +1,20 @@
+// eslint-disable-next-line import/extensions
+const config = require('10up-toolkit/config/webpack.config.js');
+
+const webpack = require('webpack');
+
+const WORDPRESS_NAMESPACE = '@wordpress/';
+const BUNDLED_PACKAGES = ['@wordpress/icons', '@wordpress/interface', '@wordpress/style-engine'];
```
nicholasio, line 7: I'm not sure why this is needed. Are we essentially leaving these packages alone and not externalizing them? Meaning that we're bundling these three packages in our bundle?

nicholasio:
> @nicholasio Just tested this by trying to only import the `IsAdmin` component like so:
> 
> ```js
> import { IsAdmin } from 'fabiankaegy-block-components-fork';
> 
> export default function BlockEdit() {
> 	return (<IsAdmin><p>test</p></IsAdmin>)
> }
> ```
> 
> and it does not do any treeshaking.
> 
> You can test the current transpiled version like so:
> 
> ```shell
> npm install --save fabiankaegy-block-components-fork@next
> ```

I'm not sure about three shaking, maybe we can look into this in a separate PR/issue?

webpack.config.js
fabiankaegy: These Bundled packages are taken straight from the Dependency Extraction Webpack plugin. https://github.com/WordPress/gutenberg/blob/68d492a15be91cf621e91d5aa16d916ec7c1bd44/packages/dependency-extraction-webpack-plugin/lib/util.js#L2-L6

These packages are not included in WordPress core as dependencies and therefore need to be installed via npm if you want to use them. Therefore they should also not get externalized.

package.json
fabiankaegy: Done 👍

fabiankaegy:
@nicholasio Your comments have been addressed :) I agree that we should get this in now and then work on the code splitting feature next. I've opened a follow up issue here: #97 and would love to hear your thoughts on how to best address this in there :)

nicholasio (approved):
LGTM

PR #101, opened 2022-03-16
Fix/content picker infinite loop

### LINKED ISSUE #99: ContentPicker: Maximum update depth exceeded
### Describe your question

Hi,

I try to use the Content Picker component to make a featured pages block. But I get an error:

`Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.`

I've made a repository with my code: https://github.com/huubl/featured-pages-plugin

I tried the [example block](https://github.com/10up/block-components/blob/develop/example/src/blocks/hello-world/index.js) too. But it doesn't set the `selectedPost` attribute somehow.

Building blocks and React is new to me and I can't find what I'm doing wrong. Can you point me in the right direction?

### Code of Conduct

fabiankaegy:
### Description of the Change

Fix infinite loop caused by circular dependency between two react hooks. 

Closes #99

### Verification Process

Use the example hello world block to select posts

### Checklist:

### Changelog Entry

Fixed: Infinite loop on post selection in content picker

### Credits

Props @huubl

### DIFF
--- a/components/content-picker/index.js
+++ b/components/content-picker/index.js
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types';
 import arrayMove from 'array-move';
 import styled from '@emotion/styled';
 import { select } from '@wordpress/data';
-import { useState, useEffect, useMemo } from '@wordpress/element';
+import { useMemo } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import { v4 as uuidv4 } from 'uuid';
 import { ContentSearch } from '../content-search';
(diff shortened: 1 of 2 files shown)


tlovett1:
@nicholasio can you take a look at this?

nicholasio (approved):
LGTM.

@tlovett1 I remember we chatted about this a while ago and the solution @fabiankaegy came up with is precisely what I suggested at the time.

PR #110, opened 2022-04-20
Add: `Repeater` Component

### LINKED ISSUE #108: Add: `Repeater` field Component
Repeater fields are a common pattern we see when working with blocks. A component to make it easier to work with would be great.

Sidsector9:
### Description of the Change

Closes #108 

#### Simple Repeater fields

https://user-images.githubusercontent.com/17757960/164195860-14ab36cc-c8e4-45f5-8b15-3f631bd5d6d7.mov

#### Grouped Repeater fields

https://user-images.githubusercontent.com/17757960/164619597-cfc8b81e-098d-4042-b5cd-66a88f50e422.mov

### Checklist:

### Credits

Props @Sidsector9, @fabiankaegy


components/repeater/index.js
```diff
@@ -0,0 +1,82 @@
...
+		});
+	}
+
+	/**
+	 * Updates the item currently being edited.
+	 *
+	 * @param {number} index The index at which the item should be updated.
+	 * @param {string|number|boolean} value The value that should be used to updated the item.
+	 */
+	function setItem(index, value) {
```
fabiankaegy: I'd love it if the setitem function and all the others already has the index inherited from the map and therefore the developer doesn't need to set it manually anymore :)

components/repeater/index.js
```diff
@@ -0,0 +1,82 @@
...
+			.filter((item, innerIndex) => index !== innerIndex);
+		updateBlockAttributes(clientId, { [attribute]: repeaterDataCopy });
+	}
+
+	return (
+		<>
+			{repeaterData[attribute].map((item, key) => {
+				return children(item, key, setItem, removeItem);
+			})}
+			<Button variant="primary" onClick={() => addItem()}>
```
fabiankaegy: I think we should allow developers to override the string in the button

components/repeater/index.js
```diff
@@ -0,0 +1,82 @@
...
+ * The Repeater Component.
+ *
+ * @param {object} props React props
+ * @param {Function} props.children Render prop to render the children.
+ * @param {string} props.attribute property of the block attribute that will provide data for Repeater.
+ * @returns {*} React JSX
+ */
+export const Repeater = ({ children, attribute }) => {
+	const { clientId, name } = useBlockEditContext();
+	const { updateBlockAttributes } = dispatch('core/editor');
```
fabiankaegy: Can we update these to use the exported store variable instead of the string?

So importing the store as blockEditorStore and then passing that variable instead of the core/editor string? :) just because it is more consistent and concise

fabiankaegy (commented):
Thank you so much for working in this. I've not tested it locally jet but the code is looking really good. I left some minor notes in the review :)

components/repeater/index.js
fabiankaegy: This was already addressed in your latest push :)

fabiankaegy:
Also before this pr can be approved we need to add the documentation etc :)

components/repeater/index.js
Sidsector9: I was working on it, didn't expect the draft PR would be reviewed 😅 
Added in [9b141b9](https://github.com/10up/block-components/pull/110/commits/9b141b9e0f9f0e478032ad67255efb2db2b446a4)

components/repeater/index.js
Sidsector9: I should've done this! Added in [debedc6](https://github.com/10up/block-components/pull/110/commits/debedc678c80857e9d4de9342eaa2f8522db8e84)
Thank you!

components/repeater/index.js
Sidsector9: Added in [8980d6a](https://github.com/10up/block-components/pull/110/commits/8980d6a54625ca5b557464a5066b99527f06dd67)

fabiankaegy:
@Sidsector9 This is looking really really good. 

If you have the time do you think you can also add an example block in the `/examples` folder? If you want to go above and beyond it would also be able to add some test cases to the component. You can take a look at the tests that already exist here: https://github.com/10up/block-components/blob/develop/cypress/integration/

We also need to add the component to the list of components in the main readme file.

Please let me know if you don't have time for that or are stuck with anything :)

Sidsector9:
@fabiankaegy that's great to hear!

I can work on the tests and example tomorrow 👍

Sidsector9:
Example of a grouped repeater field added in [52cf72b](https://github.com/10up/block-components/pull/110/commits/52cf72b85d7e4d54b0915654b17d252867770a9a)

Screen Shot 2022-04-21 at 2 20 13 PM

Sidsector9:
@fabiankaegy updating on the progress - the example, readme and E2E tests are pushed.

fabiankaegy (approved):
This is looking great. @tlovett1 You can do a final review but from my perspective, this is ready for merge. Even includes e2e test :)

fabiankaegy:
We talked about this ticket in our block editor chat today. One thing that came out of that discussion is that we should add the ability to make the repeated field reorderable to the component. 

```js
export function BlockEdit(props) {
    const { items } = attributes;

    return (
        <Repeater attribute="items" allowReordering={ true }>
            {( item, setItem, removeItem ) => (
                <TextControl value={item} onChange={setItem} />
            )}
        </Repeater>
    );
}
```

In order to do this, we should add a property to the `Repeater` component to opt-in to reordering. 

The block components package [already uses `dnd-kit` in the `ContentPicker`](https://github.com/10up/block-components/blob/c49f04f527f07b6ec5e31cef1884495d71cb7d7e/components/content-picker/SortableList.js#L31-L44) so we should be able to reuse that here.

Sidsector9:
@fabiankaegy I've added the sorting functionality, however I failed at writing the tests. I tried [this library](https://www.npmjs.com/package/@4tw/cypress-drag-drop) to perform drag-drop in the test but the repeater items just doesn't move for some reason. Not sure what I'm missing 🤔 Can you please take a look at it?

Below is the video for the sorting feature:

https://user-images.githubusercontent.com/17757960/173248066-c6275473-45e0-4e6b-bac1-7244a855d425.mov

components/content-picker/PickedItem.js
```diff
@@ -8,7 +8,7 @@ import { __ } from '@wordpress/i18n';
 import { useSelect } from '@wordpress/data';
 import { useEffect } from '@wordpress/element';
 
-const DragHandle = (props) => (
+export const DragHandle = (props) => (
```
fabiankaegy: Let's move this into its own component while we are at it. Makes it easier to maintain

components/repeater/index.js
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	/**
+	 * Updates the item currently being edited.
+	 *
+	 * @param {string|number|boolean} value The value that should be used to updated the item.
+	 * @param {number} index The index at which the item should be updated.
+	 */
+	function setItem(value, index) {
+		const repeaterDataCopy = JSON.parse(JSON.stringify(repeaterData));
```
fabiankaegy, line 114: Let's use spreading here since it is a more common pattern. 

```suggestion
		const repeaterDataCopy = [...repeaterData];
```

components/repeater/index.js
```diff
@@ -0,0 +1,256 @@
...
+		updateBlockAttributes(clientId, { [attribute]: repeaterDataCopy });
+	}
+
+	/**
+	 * Removes an item from the list.
+	 *
+	 * @param {number} index The index of the item that needs to be removed.
+	 */
+	function removeItem(index) {
+		const repeaterDataCopy = JSON.parse(JSON.stringify(repeaterData)).filter(
```
fabiankaegy, line 130: Same here: 
```suggestion
		const repeaterDataCopy = [...repeaterData].filter(
```

fabiankaegy (commented):
This is really cool! Thank you sooo much for working on this :) Added a few minor notes to the PR :)

components/repeater/index.js
Sidsector9: @fabiankaegy I made this change because the earlier version caused problems due to shallow-copy.
Any changes made to the objects in the `repeaterDataCopy` array were also reflected in `repeaterData`.
What is your opinion on using [structeredClone()](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone)?

fabiankaegy: @Sidsector9 okay that makes sense then :) maybe just add a code comment on here so it is easier to understand why this was implemented that way so it is easier to maintain :)

components/content-picker/PickedItem.js
Sidsector9: Will this require its own README?

fabiankaegy: No I don't think so. If we ever expand it in the future maybe but not right now.

fabiankaegy:
I just merged `develop` into this branch in order to full in the latest fix for the tests so that they should now pass again

fabiankaegy (approved):
This looks great from the code perspective and also works as expected 👍 

@Sidsector9 Thanks for your hard work.

PR #115, opened 2022-06-16
Feature: Query filter

Antonio-Laguna:
### Description of the Change

This PR introduces a new prop on both `ContentSearch` and `ContentPicker`. 

`queryFilter` can be passed to `ContentSearch` and it's passed from `ContentPicker` to `ContentSearch`. This parameter needs to be a `function` that gets they query as a parameter and that must return a string which will become the query to perform.

By default, the query is just returned so it's backwards compatible.

### Checklist:

- [X] I have read the [**CONTRIBUTING**](https://github.com/10up/.github/blob/trunk/CONTRIBUTING.md) document.
- [X] My code follows the code style of this project.
- [X] My change requires a change to the documentation.
- [X] All new and existing tests passed.

### Changelog Entry

> Added - `queryFilter` has been added to both `ContentSearch` and `ContentPicker` allowing you to pass a new function that modifies the query before it's performed.


components/content-picker/index.js
```diff
@@ -186,12 +190,13 @@ ContentPicker.propTypes = {
 	contentTypes: PropTypes.array,
 	content: PropTypes.array,
 	placeholder: PropTypes.string,
-	mode: PropTypes.string,
+	mode: PropTypes.oneOf(['post', 'user', 'term']),
```
Antonio-Laguna, line 193: Here and below, I think this offers more help with IntelliSense and validation in general.

components/content-search/index.js
```diff
@@ -69,10 +77,9 @@ const ContentSearch = ({ onSelectItem, placeholder, label, contentTypes, mode, p
 					break;
 			}
 
-			return searchQuery;
+			return queryFilter(searchQuery);
 		},
-		// eslint-disable-next-line react-hooks/exhaustive-deps
-		[perPage, contentTypes],
+		[perPage, contentTypes, mode, queryFilter],
```
Antonio-Laguna, line 82: `mode` should also retrigger a new callback here otherwise it's not a reactive prop. It's unlikely to change in any case but it could have.

fabiankaegy (approved):
Looks good to me 👍  Thanks for the addition

PR #116, opened 2022-06-20
add prepublish only script to ensure that we always publish build changes

fabiankaegy:
This is to ensure that we always deploy the build files.


Antonio-Laguna (approved):
This LGTM

PR #124, opened 2022-07-08
Add `fetchInitialResults` prop to `ContentSearch`

### LINKED ISSUE #43: Content Picker: Support showing an initial set of results on focus
Many content picking scenarios are attempting to select from the latest relevant items. It isn't transparent to the user what to expect once they start typing and they may not even need to search at all.

Surface an initial X set of results on focus, defaulting to reverse chronological of whatever content is included in search, but I imagine all of this would be configurable via props. IMO it should default to showing those results, with a prop to not show them.

**Designs**
n/a

**Describe alternatives you've considered**
Not an alternative but could consider only showing initial results on focus if nothing has been selected in that picker yet. I am not strongly opinionated on this point at the moment.

**Additional context**
This has been a direct request on other projects and I agree that it is better "starting point" UX. It is not equivalent to core linking, because in that situation there is no real predicting whether a user is trying to make an external or internal content link.

xipasduarte:
### Description of the Change

This new prop allows control over initial state for the queries in `ContentSearch`. It
will trigger a fetch when mounting the component and cache the response for display by focusing the search field, making the user aware of what they might be selecting from.

A new state was added to keep track if the search field is focused. This allows for better control of the results area display. To keep previous expectations the default value is `false`.

Then, on the `ContentPicker` component extend the same prop with a default value of `true` to enhance UX.

Closes #43

### Alternate Designs

Another alternative was considered for the UX on `ContentPicker`, but the logic is fairly simple to fix if want to. See the issue for more details.

### Possible Drawbacks

I thought about adding an option for the user to specify if they want this fetch to happen on every item count and the effect it would have. I believe that this other prop could be added later without any drawbacks. The only issue might be from a UX stand point, if it becomes preferable to have `fetchInitialResults` for all item counts in `ContentPicker`.

### Verification Process

- Manual testing using block `example/hello-world` which has a post picker following the expected scenarios:
  - When the search field is focused with `fetchInitialResults = true` the results of the query should be displayed.
  - When the search field is focused with `fetchInitialResults = false` the current behavior should occur, with fetch.
  - When the `ContentPicker` is used it should fetch initial results when no items exist `content.length === 0`. Otherwise the current behavior is expected.
- By making `fetchInitialResults = false` as a default the current working logic is kept for `ContentSearch`.

### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://github.com/10up/.github/blob/trunk/CONTRIBUTING.md) document.

### Changelog Entry

> Added - New prop to `ContentSearch` to indicate that results should be fetched when on mount (`fetchInitialResults`).
> Added - Initial fetch for `ContentPicker`. It should fetch results if no items were selected yet.

### Credits
This new prop allows initial results to be fetched when mounting `ContentSearch`. These results will display when focusing the search field, making the user aware of what they might be selecting from.

To keep previous expectations the default value is `false` on `ContentSearch`.
Props @helen for detailing an initial approach to solve this UX problem at #43.


Antonio-Laguna:
Hey @xipasduarte, firstly I want to thank you for taking the time for both, opening a PR and also taking the time to go over and clean some code! Thanks! 

I'm not sure if Helen's intention was to actually perform the query on focus or rather show some results on focus. 

I've recently participated in a project which has over 200,000 published articles in which _any_ search took over 10 seconds which makes me wary of adding this behavior since based on payload size and network conditions, the UI can look unresponsive for no apparent reason. 

While I understand this is very much an edge case, I'd say this should just prompt the search on mount so results are ready for whenever the user decides to focus the input, it just gets the results. (this would mean a rename on the new prop). I think it's kind of safe to assume that, if the block is selected (so this component is mounted), you're _likely_ to perform actions that might lead to a search that would benefit from this new prop and behavior. 

What are your thoughts on this? 

Also interested to know about the opinions of @fabiankaegy and @tlovett1 on this!

Thanks again!

xipasduarte:
Hi @Antonio-Laguna, I'm happy to help and hope I can keep at it 🙂.

I agree that there is an expectation of using the component once it is selected. We also have a focus on the search field when adding the block, which enforces this expectation. So, regardless of the edge case, I think it's a great idea to "warm up" with the initial results rather than fetching on focus.

Any suggestion for the prop rename? Maybe something like `fetchInitialResults`?

I'll leave some time for others to give feedback before going forward with any changes. I'm going to just test and see if the presence of the search field, even if empty, is affecting the performance of the query. I expected this would be like fetching the first page of any post type archive, but if there are issues we can remove the query arg `search` and benefit in the edge cases.

Antonio-Laguna:
@xipasduarte I think `fetchInitialResults`  is a good name!

xipasduarte:
I made the changes to consider `fetchInitialResults` as the new prop and the expected behavior. 

From what I could gather the first fetch will be an empty search field, so this means it is analogous to the first page of the archive and should be fast. (If anyone has a particularly hard example, please run these with it to double check.) Given this, should we expose this prop in the `ContentPicker` as well? I don't see a lot of harm in doing so, just an added option to consider. As a default it should be `true` to offer the best UX (which is fetching in most cases).

components/content-search/index.js
```diff
@@ -285,6 +293,7 @@ const ContentSearch = ({
 
 	const hasSearchString = !!searchString.length;
 	const hasSearchResults = searchResults && !!searchResults.length;
+	const hasInitalResults = fetchInitialResults && isFocused;
```
Antonio-Laguna: Minor typo here 

```suggestion
	const hasInitialResults = fetchInitialResults && isFocused;
```

Antonio-Laguna:
Hey @xipasduarte ! Thanks for doing this again! I've flagged a minor typo but other than that it's looking good! I agree with it being added to `ContentPicker` as well. Most props end up being in both. Also agree with being `true` as it's mostly harmless in these scenarios and enhances UX as you state!

xipasduarte:
@Antonio-Laguna I've made the correction and extended the `fetchInitialResults` prop to the `ContentPicker` componente. I also had a look on the description to make things more in line with what we ended up doing.

Is it possible to enable the tests to run so as to see if everything passes?

Let me know if anything else needs a little more work 🙂

fabiankaegy (approved):
Thank you @xipasduarte for your patience on this and for working on it in the first place. I just tested this and made one small update to not have it enabled by default. Besides that this is ready to go and approved from my end :)

PR #126, opened 2022-07-12
Add `prop` `pickedOrder` for new selected initial order

### LINKED ISSUE #125: `ContentPicker` new `prop` to control selected initial order

When using the picker my desired list output requires that the new item is pushed to the bottom of the list. In a small list this is not a problem to manually resort, also it's not too hard to redo the actual selection with `onPickChange`, but it might be simpler if we do it this way.

### Designs

_No response_

I've considered letting the user resort manually, but it seems to be too troublesome for large lists. another alternative, which I have in place, was to use `onPickChange` and redo the sort, considering it pushes to the top, but it feels like I'll be doing this every time I need this behavior change.

### Code of Conduct

xipasduarte:
When selecting a new item for the list, the only option was to have the
item pushed to the top. This adds support for two other sorts:
- `end` pushes the item to end of the `content` array
- `function` if the user defines a function they may control where to push
  the new item
- `start` becomes the default value, keeping the current behavior

### Description of the Change

When selecting a new item for the list, depending on the value of `pickedOrder` new item is pushed on the array of `content` according to its value.
- `start` becomes the default value, keeping the current behavior
- `end` pushes the item to end of the `content` array
- `function` if user defined a function they may control where to push
  the new item

As mentioned, keeping the default value at `start` avoids breaking the current behavior.

Closes #125 

### Alternate Designs

I considered only using a boolean value alone with a prop like `pickedOrderEnd`, only to negate the current behavior when needed, but the current approach seems more versatile. Also, adding the option for a function allows the user reach for control of how this is done in some specific cases (e.g. always pushing in alphabetical order).

### Possible Drawbacks

Given the default behavior is kept, I don't see any at the moment.

### Verification Process

I've only done manual testing, but am hoping to open this for feedback, especially for the prop name (`propInitialOrder`, maybe?).
Will do some unit testing/e2e depending on how #124 is reviewed (they change the same components and it would be nice to reduce conflicts).

### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://github.com/10up/.github/blob/trunk/CONTRIBUTING.md) document.

### Credits

Props @xipasduarte


fabiankaegy:
Closing this as stale for now

PR #127, opened 2022-07-14
add support for passing multiple blocks to the registerBlockExtension API

fabiankaegy:
### Description of the Change

Before this, it was only possible to add a block extension to one block at a time. This is very limiting since many of the extensions we see on a day-to-day basis are getting applied to multiple blocks simultaneously.

With this change, the first parameter of the `registerBlockExtension` accepts either a string or an array of strings.

### How to test the Change

Update example registerBlockExtension to include more than one block.

### Changelog Entry

- Added ability to provide multiple blocks to registerBlockExtension API

### Checklist:

### DIFF
--- a/api/register-block-extension/index.js
+++ b/api/register-block-extension/index.js
@@ -39,6 +39,8 @@ function registerBlockExtension(
 		return blockType === blockName;
 	};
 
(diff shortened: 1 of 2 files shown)


Antonio-Laguna (approved):
Love this! Thanks for adding this!

PR #132, opened 2022-07-22
add `useBlockParentAttributes` hook

fabiankaegy:
### Description of the Change

Sometimes there are instances where one block needs to update the attributes of another block. This especially happens with direct parent child block relationships like for example the core column block needing to update the vertical alignment value of the core columns block. 

This has always been possible with the usage of the Data API. But it is a lot of code that could be abstracted away. This is what the `useBlockParentAttributes` hook does. 

```js
import { useBlockParentAttributes } from '@10up/block-components';

function BlockEdit(props) {
    const [parentAttributes, setParentAttributes] = useBlockParentAttributes();

    return (
        <RichText
            tagName="h2"
            value={ parentAttributes.title }
            onChange={ value => setParentAttributes({ title: value }) }
         />
    );
}
```

### Changelog Entry

> Added - `useBlockParentAttributes` hook to easily interface with the attributes of the direct parent block

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ A collection of components built to be used in the block editor. These component
 - [useIcons](./hooks/use-icons/)
 - [useMedia](./hooks/use-media/)
 - [useRequestData](./hooks/use-request-data/)
+- [useBlockParentAttributes](./hooks/use-block-parent-attributes/)
 
 ## Stores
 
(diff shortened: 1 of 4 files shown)


Antonio-Laguna (approved):
This looks good to me! Very useful indeed!

PR #131, opened 2022-07-21
add `useCharacterLimit` hook

fabiankaegy:
### Description of the Change

Add a new `useCharacterLimit` hook that allows developers to define a limit for any text field. 

### How to test the Change

Use the Character Limit block in the examples folder

### Changelog Entry

> Added - `useCharacterLimit` hook to easily limit the amount of characters editors should provide

### Credits

Props @darrenjacoby

### Checklist:

- [x] I agree to follow this project's [**Code of Conduct**](https://github.com/10up/.github/blob/trunk/CODE_OF_CONDUCT.md).
- [x] I have updated the documentation accordingly.
- [x] I have added tests to cover my change.
- [x] All new and existing tests pass.

PR #133, opened 2022-07-22
add `RichTextCharacterLimit` component

darrenjacoby:
### Description of the Change
Extend `RichText` with `RichTextCharacterLimit` in order to enable character limits on rich text content. 

### How to test the Change

```js
import { RichTextCharacterLimit } from '@10up/block-components';

function BlockEdit(props) {
    const { title } = attributes;

    return (
        <RichTextCharacterLimit
            limit={30}
            enforce={true}
            tagName='h2'
            value={title}
            onChange={(title) => setAttributes({title})}
            allowedFormats={[
                'core/bold',
                'core/link'
            ]}
        />
    )
}
```

Attribute `title` should never exceed the character limit unless `enforce` is set to `false`. If `enforce` is set to `false`, the author will be able to exceed the character limit. Both instances will show a character counter below the `RichText` component.

### Credits
Props @fabiankaegy 

### Related

https://github.com/10up/block-components/pull/131 
@fabiankaegy should we favour a component versus a hook? The implementation and DX feels cleaner.

### Checklist:

_Please let me know if I can add tests._

### DIFF
--- a/components/index.js
+++ b/components/index.js
@@ -12,3 +12,4 @@ export { Repeater } from './repeater';
 export { Link } from './link';
 export { MediaToolbar } from './media-toolbar';
 export { Image } from './image';
+export { RichTextCharacterLimit } from './rich-text-character-limit';
--- a/components/rich-text-character-limit/index.js
+++ b/components/rich-text-character-limit/index.js
@@ -60,7 +60,7 @@ const RichTextCharacterLimit = (props) => {
 		const isOverLimit = getCharacterCount(str) > limit;
 
 		if (isOverLimit && enforce) {
-			// Hack which fixes an issue with `<RichText>` not updating.
+			// Workaround which fixes an issue with `<RichText>` not updating.
 			setRichTextValue('');
 			return remove(richTextContent, limit, getCharacterCount(str));
 		}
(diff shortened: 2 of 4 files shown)


darrenjacoby:
I wasn't sure if some default styling for `<Counter />` is preferable, but let me know and I can add some.

fabiankaegy:
@darrenjacoby Yeah honestly that is the main thing I am currently still thinking about. Ideally, the element should be absolutely positioned/floating in the corner of the `RichText` component. And only visible if the block is currently selected. 

I'm wondering whether we can apply some magic js based positioning to it since I don't want to introduce a wrapper element that would change the editors markup compared to the frontend.

fabiankaegy:
Maybe we can use https://floating-ui.com/ to get the element so float in the corner when the block is selected. 

We can get the selection state info via the `useBlockEditContext` hook

```js
const { clientId } = useBlockEditContext();
```

darrenjacoby:
Sounds good, I'll see what we can do based on that recommendation!

fabiankaegy:
I'm going to merge this onto a new branch in this repo so I can also add to it from there 👍

PR #134, opened 2022-07-22
Add `CharacterLimitRichText` component

fabiankaegy:
### Description of the Change

The PR adds a new Component that wraps the WordPress `RichText` component and adds a character limit. 

https://user-images.githubusercontent.com/20684594/180438859-a8c1448c-c905-4d71-aae3-ae8afd44980c.mp4

Continuation of #133 

Closes #

### Changelog Entry

> Added - `CharacterLimitRichText` component to allow greater control over specific character limits of RichText fields.

### Credits

Props @darrenjacoby @fabiankaegy 

### Checklist:

### DIFF
--- a/components/index.js
+++ b/components/index.js
@@ -12,3 +12,4 @@ export { Repeater } from './repeater';
 export { Link } from './link';
 export { MediaToolbar } from './media-toolbar';
 export { Image } from './image';
+export { RichTextCharacterLimit } from './rich-text-character-limit';
--- a/components/index.js
+++ b/components/index.js
@@ -12,4 +12,9 @@ export { Repeater } from './repeater';
 export { Link } from './link';
 export { MediaToolbar } from './media-toolbar';
 export { Image } from './image';
-export { RichTextCharacterLimit } from './rich-text-character-limit';
+export {
+	RichTextCharacterLimit,
+	getCharacterCount,
+	CircularProgressBar,
+	Counter,
+} from './rich-text-character-limit';
--- a/components/rich-text-character-limit/index.js
+++ b/components/rich-text-character-limit/index.js
@@ -29,6 +29,7 @@ const getCharacterCount = (str) => {
  */
 const Counter = forwardRef((props, ref) => {
 	const { count, limit } = props;
+	const percentage = (count / limit) * 100;
 	return (
 		<StyledCounter
 			className={cx('tenup--block-components__character-count', {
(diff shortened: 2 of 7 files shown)


fabiankaegy:
I'm not sure yet how the Circular Progress Indicator should handle showing its status when the component does not enforce the character limit. 

I'd like to do something similar to Twitter where it changes color and has some nice UX hints.

xavortm:
A suggestion for the indication:
image

And on dark background: 
image

On colored:
Screenshot 2022-07-25 at 08 58 02

The reasoning for what I did:

- The icon instead of color due to color blindness — we need a shape to indicate the status.
- I placed them on the left so that it is always visible where you start typing. See this example of the problem when the indicator is on the right: https://share.cleanshot.com/GlSDkNflBqS3nyRqUqA9
- I used solid icon for the over-the-limit state to make it easier to scan / find quickly.

And for the feature set, I think it would be ideal if we could have a Soft and Hard limits (bool) - a soft limit only counts the letters and reports, while a hard limit stops you from saving the post if it's over the required character limit.

And maybe we can use the middle example for the soft limit when over the character and the third icon for hard limit. And both to show up only after you go over the limit

darrenjacoby:
Hey @xavortm ,

There is the option for soft and hard limits, you can use `enforce={false}` on the component to use soft limits.

fabiankaegy:
Hey @xavortm 👋 

Love that design exploration. For insight the reason why I had placed the character count panel on the right hand side it that the block appender most of the time is located at the bottom. And I wanted to avoid any awkward placement. 

Also it felt a little more familiar since that is how I was used to it from Twitter. 

However I do agree that it is not ideal because unlike twitter we don't have a clear boundary around the text area but instead just have floating text. 

So I am cool with playing with the positioning. 

As for the icons I think that is another great callout that we should include. We'll have to play a little with the various states when the component is set to enforce the limit vs when the component is set to not enforce. 

On that note I'd also like to get some more insight into whether we should actually cut off the text if the user goes over the limit and prevent them from typing or whether we should handle it like you have shown in the example where the user can still type but not save the post. 
I actually had it working that way initially but without the warning indicator the UX was lacking a bit. So I think we should make it so that the user should still be able to type more but the error icon should appear and stay present even when the block is not selected. Also the saving of the editor should be locked until the text is cut shorter. 

The reason for this is that you as an editor may want to finish your thought before you start to reword / refactor your sentence to fit into the character limit. I know that's always how I do it with twitter. 

The only thing about the designs you @xavortm have shown that I would change is that the circle should be rotated like it currently is in the live example where the start and end point is at the top instead of on the right.

Would love to hear how everyone feels about these suggestions before we can make the updates and land this :)

PR #158, opened 2022-10-16
Move `Counter` to a separate component

### LINKED ISSUE #156: `Counter`: Spin off into individual component, a11y, and icon

I'm using a customized version of the `Counter` component that is currently part of the [rich text character limit](https://github.com/10up/block-components/tree/develop/components/rich-text-character-limit).

### Break out `Counter` 
I'm only using the counter in this case, not the entire component. Since I had to add it to an existing control that we specifically don't want to be `RichText`, it may be useful to spin out the `Counter` so it can be used in other contexts.

### a11y Updates
For accessibility purposes, there should probably be a couple of changes in particular:
- The warning icon could use some additional contrast since yellow on white tends to be hard to see. In my customized version, I'm using a yellow icon with a black fill.[1]
- We should look into using something like `meter` to semantically let the user know the value of characters remaining

### Icon updates
In the customized version, we also have added a "good" icon that gives the user extra information that the current value is well within the specified range.[2]

### Designs

1. Screenshot from 2022-09-30 09-00-02
2. Screenshot from 2022-09-30 08-38-16

_No response_

### Code of Conduct

Sidsector9:
## Note

- We need the modified green counter SVG with the tick added in it.

### Description of the Change

- This PR moves the Counter and Circular Progress Bar to a separate component.
- Adds a black fill to the exclamation mark of the warning icon.
- Adds red color to icon when text goes beyond limit.

[WordPress Colors](https://make.wordpress.org/design/handbook/design-guide/foundations/colors/) are used for reference.

Closes #156 

### Screenshot

Screenshot 2022-10-16 at 1 18 42 PM

Screenshot 2022-10-16 at 1 19 52 PM

Screenshot 2022-10-16 at 1 22 12 PM

### Changelog Entry

> Changed: yellow color of the warning icon
> Changed: fill of the exclamation to black color
> Changed: icon color to red when text goes beyond limit

### Credits

Props @Firestorm980 @Sidsector9 

### Checklist:


jeffpaul:
@cadic FYI that this will likely be helpful to you on the related OSS project you're looking into

example/src/blocks/rich-text-character-limit/index.js
```diff
@@ -1,7 +1,7 @@
 import { registerBlockType } from '@wordpress/blocks';
 import { __ } from '@wordpress/i18n';
 import { useBlockProps } from '@wordpress/block-editor';
-import { RichTextCharacterLimit } from '@10up/block-components';
+import { RichTextCharacterLimit } from '../../../../components/rich-text-character-limit';
```
fabiankaegy: @Sidsector9 What's the reason for this change here? I don't think we should use the relative file path here.

fabiankaegy (changes requested):
This is great :) I only left one comment about the example code.

Otherwise this is ready to go 🚀

example/src/blocks/rich-text-character-limit/index.js
Sidsector9: Now that you ask, I can't recollect why I did this 🤔 

I fixed it in [45fc61f](https://github.com/10up/block-components/pull/158/commits/45fc61fc63b001b59d8ad2cbd05ae589632c85fe)

FYI, as per the design, there's work pending on the green icon. A tick inside the circle is missing which needs to be added.

fabiankaegy (approved):

PR #162, opened 2022-10-25
Enhancement: Add `renderItemType` prop to `ContentSearch`

### LINKED ISSUE #160: Customize the `ContentSearch` item type label

I'm using the `ContentSearch` component to allow selection of a single post (of any public post type).

When using it only with post types, the item type label can feel a bit misleading - from what I can see, the item type label is set to the "mode" property of the item, so in this case it will be `post` as the content type is a post.

However, as I'm using it only for post types, when searching for a page, or CPT, it feels a bit misleading being labelled as `post` (also the default post type).

It would be great to be able to pass through a custom property (of the item in the list) instead, in this case the actual post type.

### Designs

Before:
image

After:
image

_No response_

### Code of Conduct

rmorse:
### Description of the Change

Adds a callback to allow customisation of the item type in the suggestions list.

Closes #160 

### How to test the Change

Add the `renderItemType` prop to the `ContentSearch` component.
```jsx
<ContentSearch
	{ ...props }
	renderItemType={ ( suggestion ) => {
		return suggestion.subtype;
	} }
/>
```

### Changelog Entry

> Added - New feature - customise the item type rendered in the ContentSearch component.

### Checklist:

_I had trouble creating the dev environment so couldn't spin up the tests._

### DIFF
--- a/components/content-search/SearchItem.js
+++ b/components/content-search/SearchItem.js
@@ -49,13 +49,6 @@ const SearchItem = ({
 	const textContent = getTextContent(richTextContent);
 	const titleContent = decodeEntities(textContent);
 
-	// Rename 'post_tag' to 'tag'. Ideally, the API would return the localised CPT or taxonomy label.
-	let type = suggestion.type === 'post_tag' ? 'tag' : suggestion.type;
-	// Allow type to be overriden by renderType
-	if (renderType) {
-		type = renderType(suggestion);
-	}
-
 	return (
 		<Tooltip text={decodeEntities(suggestion.title)}>
 			<ButtonStyled
@@ -89,7 +82,9 @@ const SearchItem = ({
 					</span>
 				</span>
 				{showType && (
-					<span className="block-editor-link-control__search-item-type">{type}</span>
+					<span className="block-editor-link-control__search-item-type">
+						{renderType(suggestion)}
+					</span>
 				)}
 			</ButtonStyled>
 		</Tooltip>
@@ -100,7 +95,10 @@ SearchItem.defaultProps = {
 	id: '',
 	searchTerm: '',
 	isSelected: false,
-	renderType: null,
+	renderType: (suggestion) => {
+		// Rename 'post_tag' to 'tag'. Ideally, the API would return the localised CPT or taxonomy label.
+		return suggestion.type === 'post_tag' ? 'tag' : suggestion.type;
+	},
 };
 
 SearchItem.propTypes = {
--- a/components/content-search/SearchItem.js
+++ b/components/content-search/SearchItem.js
@@ -31,14 +31,30 @@ const ButtonStyled = styled(Button)`
  * @param {string} props.searchTerm the search term
  * @param {boolean} props.isSelected whether the item is selected
  * @param {string} props.id the id of the item
+ * @param {Function} props.renderType a callback to override the type text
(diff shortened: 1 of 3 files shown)


fabiankaegy:
Thank you for working on this @rmorse 🚀 

I do like this enhancement quite a bit! There is one thing that I want to think through before merging this in though. I'm wondering whether it would be a better more flexible/maintainable API to provide a way for consumers of this component to pass in their own `suggestionTemplate`.

I've seen that pattern get used by various libraries like for example the [alphagov-accessible-autocomplete](https://github.com/alphagov/accessible-autocomplete#templates-default-undefined)

That way we could ensure that we can support all kinds of customizations. 

The reason why I am suggesting that we discuss this is that I don't want to end up in a place where we need 5 different functions to modify each small little piece of the suggestion individually because it can get very messy to maintain. 

I'm very open to suggestions though and maybe in this case one doesn't mean the other isn't also useful since changing the type label is a more common action and I could see how having to rebuild the entire suggestion markup just to modify this one piece can be overkill. 

I'd love to get @Antonio-Laguna to chime in here aswell how he thinks we could proceed with this.

rmorse:
Thanks Fabian, I just rolled with this for now as it was something I needed (immediately!) and thought I'd share it in case it's something you wanted to add in.

Totally understand the need to future proof this and not add too many props.

I do like the idea of the suggestionTemplate, sounds like a great idea for complete customisation

components/content-search/SearchItem.js
```diff
@@ -87,6 +100,7 @@ SearchItem.defaultProps = {
 	id: '',
 	searchTerm: '',
 	isSelected: false,
+	renderType: null,
```
Antonio-Laguna: @rmorse thoughts on putting the logic you added on L53 here as a sensible default? That way the logic from L52:L57 can be simplified as `renderType` would always have a function (unless the prop has a value type it shouldn't but that's ok). 

If you think it's right you should also update the default value on `content-search/index.js` so it's `undefined` instead of `null`

Antonio-Laguna (approved):
@rmorse thanks a ton for working on this!

As @fabiankaegy hints, I think the way to go moving forward is to have template parts as props. This is something I've been pondering in general for the blocks as I've seen the need for complex ones such as `ContentPicker`. 

I've reviewed this (nice work!) and I do see the value of this prop vs the more cumbersome path of having to write your own component which is a bit overkill. When we change to templates as props this will just be a prop too and you can do (or not do) as you want with the prop.

👍 from me!

components/content-search/SearchItem.js
rmorse: @Antonio-Laguna do you mean something like this?

```
SearchItem.defaultProps = {
	...
	renderType: ( suggestion ) => {
		return suggestion.type === 'post_tag' ? 'tag' : suggestion.type;
	},
	...
}
```

Agreed, this looks cleaner to me, happy to add in.

Antonio-Laguna: Exactly that!

rmorse: Great stuff, will get this tidied up tomorrow

rmorse: @Antonio-Laguna update just landed

fabiankaegy:
Thanks again for the update here :) This is a great addition!

PR #163, opened 2022-10-25
Add a way to supply a fallback image to the `Image` component

### LINKED ISSUE #161: Add the ability to provide a fallback image to the `Image` component that gets used in the block `example`
Currently, since the `Image` component relies on the `imageId` there is no way to provide a fallback / example image. This leads to a not ideal experience when using block patterns / the block preview in general.

It would be great to provide a way to solve this issue.

fabiankaegy:
### Description of the Change

This is an exploration of how we could potentially handle adding a fallback image option to the image component.

Closes #161 

### Changelog Entry

> Added - Ability to provide a fallback image for the `Image` component

### Credits

Props @ncoetzer @sudar @ajmaurya99 

### Checklist:


example/src/blocks/image-example/block.json
```diff
@@ -5,13 +5,24 @@
...
     "attributes": {
-        "imageId": {
-            "type": "number"
+        "image": {
+            "type": "object",
+			"default": {
+				"id": null,
+				"fallback": null
+			}
         },
```
ncoetzer, line 26: @fabiankaegy Essentially, I think we could have gone two ways about this:

1. The way you approached it here by using a single prop structured as an object with two values, i.e., `id` & `fallback`, respectively. However, I'm just wondering about any possible confusion about what type of data needs to be provided for these two values, i.e., number vs. string.
2. **OR** perhaps it would be more clear (and less complex) to define the fallback image URL as a separate prop (optional), e.g.,  
```
"imageId": {
    "type": "number",
    "default": null
},
"fallbackImageUrl": {
    "type": "string",
    "default": ""
}

fabiankaegy:
Closing this as stale for now

PR #166, opened 2022-11-07
Update readme.md example of link component

adimitrovsf:
Repalces 'text' with 'linkText' in props change; I was reusing the code example here and didn't notice that I wasn't updating the component props when editing the link in the editor instead of the popup.


fabiankaegy (approved):
Thanks for the fix :)

PR #172, opened 2022-11-11
Add ability to show inline media controls for the `Image` component

### LINKED ISSUE #164: Add a new `hasInlineControls` option to `Image` component
There often are instances where we either need to have multiple images in one block or where the toolbar replaces flow isn't ideal. For those cases, it would be great to also offer an inline solution for replacing and or removing images. 

The idea here would be to add a new optional property to the `Image` component which is a boolean to enable this additional behavior. 

When this boolean is present the block will show inline controls when the block is currently selected. These controls should be absolutely positioned so that they don't create any sort of layout shift.

These controls should also take into account whether the image is marked as optional or not. If it is optional the remove button should also be shown inline. Otherwise only the replace button should be shown.

ncoetzer:
### Description of the Change

Sometimes, we have blocks that require multiple images, and this change aims at offering an inline solution for replacing and or removing images where the toolbar replaces flow isn't ideal. 

Closes https://github.com/10up/block-components/issues/164

#### Steps to test
- Add the _Multiple Image Example_ block to the page.  
image

- You should see media placeholders for 3 images.  
image

- Once you have set an image, you should see the "Replace" and "Remove" icon buttons when hovering over the image. This is because the new (optional) `hasInlineControls` props have been set to `true`.  
image  

**_Note_**: If text is preferred over icons, this could be easily replaced. We should probably get UX/Design feedback on the icons. 

- You should be able to replace the image with another and see the updated image display (updating the image id in the DB)
- The "Remove" icon button should not be visible if the `isOptional` props have been set to `false`.  
image

##### Optional
```
<Image 
  id={imageId} 
  size="large" 
  onSelect={(image) => setAttributes({imageId: image.id })} 
  className="example-image" 
  focalPoint={focalPoint} 
  onChangeFocalPoint={(value) => setAttributes({focalPoint: value})} 
  onRemove={() => setAttributes({imageId: null})} 
  hasInlineControls={true} 
/>
```
##### Required
```
<Image 
  id={imageId} 
  size="large" 
  onSelect={(image) => setAttributes({imageId: image.id })} 
  className="example-image" 
  focalPoint={focalPoint} 
  onChangeFocalPoint={(value) => setAttributes({focalPoint: value})} 
  onRemove={() => setAttributes({imageId: null})}
  hasInlineControls={true} 
  isOptional={false}
/>
```

### Changelog Entry

> Added - Ability to show inline media controls for the `Image` component

### Credits

Props @fabiankaegy 

### Checklist:


example/src/blocks/multiple-image-example/edit.js
```diff
@@ -0,0 +1,25 @@
...
+        setAttributes
+    } = props;
+
+    const { image1, image2, image3, focalPoint1, focalPoint2, focalPoint3 } = attributes;
+    const blockProps = useBlockProps();
+
+    return (
+        <div {...blockProps}>
+
+            <Image id={image1} size="large" onSelect={(image) => setAttributes({image1: image.id })} className="example-image" focalPoint={focalPoint1} onChangeFocalPoint={(value) => setAttributes({focalPoint1: value})} hasInlineControls={true} onRemove={() => setAttributes({image1: null})} isOptional={false} />
```
fabiankaegy: Could you refactor this to have linebreaks between the individual attributes? That would make it much more readable :)

fabiankaegy (changes requested):
Thank you for working on this @ncoetzer. This will be such a massive improvement for the image component. 

I had asked @xavortm to mock up some designs and think through some edge cases and he provided a few mockups here: https://github.com/10up/block-components/issues/164#issuecomment-1318309508 

I would love your thoughts on them so we can update this PR to match that design.

---


I also have one more significant concern that I am not sure how we can best address it. Something that we should always try with these components is markup parity with the frontend of the site. Which in this case means outputting an `img` tag. This way all the custom styles that get applied from the theme/block author actually get picked up by the image same as they do on the frontend. without creating additional work to stye the admin.

The way this currently works we lose this parity because we need to wrap the `img` tag in a `div` in order to absolutely position the button elements on top of the image. 

I'm not sure what a good solution for this is yet. But I would like us to experiment with either moving the button elements to become siblings of the image tag and then using JS in the editor to handle the positioning similar to how Gutenberg does it with popovers etc. (Maybe we can even use the popover with an `anchorRef` and style it so it looks like it is inline.)

Let me know if you want to brainstorm on this together :) Also don't feel the need to sync time into this. I'm also happy to pick this up from here if you don't have any more time to spend on this :)

fabiankaegy:
Hey @ncoetzer 👋 


https://github.com/10up/block-components/assets/20684594/32fcc007-6c79-4587-ba91-bba245c61946


The new look you shared for this component is sooooo nice! I really love it! :) Great work! 

The only thing I still struggle with, is that this change introduces a new additional `div` in the markup, which I would consider a breaking change since it might mess with existing style setups that expect this component to just render the `img` tag. 

Because of that I think we have two available options:

1. Refactor the code here to get rid of the additional component. You can look how @darrenjacoby achieved this in #238 (`useFloating` + `cloneElement`) so that the output rendered by the `Image` component still is only the `img` tag
2. Abandon this PR here and instead move this new code into the exploration started in #222

fabiankaegy:
Closing in favor of #269

PR #183, opened 2022-12-20
Doc updated: Image component Focal Picker needs default settings

### LINKED ISSUE #182: Image component Focal Picker needs default settings
### Describe the bug

For the Image component, if you wish to use the Focal Picker, you must declare default values in your block.json attributes for it to render. 

cc: @ncoetzer 

### Steps to Reproduce

1. Add component as outlined in the [readme](https://github.com/10up/block-components/blob/develop/components/image/readme.md)
2. Add your attributes to block.json
```
{
  ...
  "attributes": {
    "imageId": {
      "type": "number"
    },
    "focalPoint": {
      "type": "object"
    }
  }
}
```
See that the Focal Picker doesn't render.  

Proposed solution would be to outline in the readme file that default settings must be added for it to render.  

```
{
  ...
  "attributes": {
    "imageId": {
      "type": "number"
    },
    "focalPoint": {
      "type": "object",
      "default": {
        "x": 0.5,
        "y": 0.5
      }
    }
  }
}
```

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

faisal-alvi:
### Description of the Change
This PR simply adds a note in the readme file for the image block component that default settings must be added when a `focalPoint` is used. Also capitalizes the first character in the description column.

Closes #182

### How to test the Change
1. Add component as outlined in the [readme](https://github.com/10up/block-components/blob/develop/components/image/readme.md)
2. Add your attributes to block.json
```
{
  ...
  "attributes": {
    "imageId": {
      "type": "number"
    },
    "focalPoint": {
      "type": "object"
    }
  }
}
```
3. See that the Focal Picker should render.

### Credits
Props @colinswinney @faisal-alvi 

### Checklist:

### DIFF
--- a/components/image/readme.md
+++ b/components/image/readme.md
@@ -42,9 +42,9 @@ function BlockEdit(props) {
 
 | Name       | Type              | Default  |  Description                                                   |
 | ---------- | ----------------- | -------- | -------------------------------------------------------------- |
-| `id` | `number`    | `null`   | image id          |
+| `id` | `number`    | `null`   | Image ID          |
 | `onSelect` | `Function` | `null` | Callback that gets called with the new image when one is selected |
-| `size` | `string` | `large` | name of the image size to be displayed |
-| `focalPoint` | `object` | `undefined` | optional focal point object |
+| `size` | `string` | `large` | Name of the image size to be displayed |
+| `focalPoint` | `object` | `undefined` | Optional focal point object. Default settings must be added when used. |
 | `onChangeFocalPoint` | `function` | `undefined` | Callback that gets called with the new focal point when it changes |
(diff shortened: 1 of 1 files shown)


fabiankaegy (approved):
Thanks for these updates :) 

I think we can merge this in as is. As a follow up I would like to actually change the implementation in the `Image` component to automatically have a fallback value for the focal point in place so that this is no longer an issue.

faisal-alvi:
That is exactly what I thought earlier but then it seems like we also need to handle the `onChangeFocalPoint`'s fallback function. So I skipped it and just changed the readme. :)

fabiankaegy:
@faisal-alvi I created #184

PR #189, opened 2023-01-15
Only output icon if one is set

tlovett1:
### Description of the Change
In UI Kit, we were just getting a spinner in the icon toolbar dropdown when none was set. This fixes that problem.

### How to test the Change
Confirm if no icon is set, there is no spinner in the toolbar.

### Changelog Entry
- Don't output icon if none is set in toolbar

### Checklist:


fabiankaegy (approved):
Looks good to me 👍

PR #191, opened 2023-01-17
Exploration: Refactor codebase to TypeScript

fabiankaegy:
### Description of the Change

This PR aims to refactor the entire codebase of the `@10up/block-components` to use [TypeScript](https://www.typescriptlang.org/). The reason for using TypeScript is to provide consumers of the Block Components package with much better IDE integration and autocomplete when using any of the components/hooks from this package. Additionally, it allows us to gain more confidence in the code of this library without relying on the current PropTypes workflow. 

### How to test the Change

All the tests should continue to pass. Additionally, you can go into the `example/src/blocks/` folder and try out the better autocomplete/IDE integration this change would entail for the library users. 

### Changelog Entry

> Changed - Refactor block components to use TypeScript internally

### Credits

Props @fabiankaegy, @Antonio-Laguna, @ravinderk

### Checklist:


fabiankaegy:
Hey @Antonio-Laguna @nicholasio @tlovett1 

I would love to get your thoughts on this exploration. My goal is to ship better IDE integration and autocomplete for anyone using the Block Components. But I'm unsure whether using TypeScript would hinder more folks from contributing to the library. 

I would consider myself to be a TypeScript noob (as you can probably quickly tell by the code in this PR  😂 ). But even with the naive implementation here it took me a few hours to convert the entire codebase over to TS, and I didn't need to change much. The few things I did change are actually potential bugs anyways so the changes improve the stability of the library itself. 

I would love to get your all's thoughts on the topic. 

Thanks in advance :) 

(And just to be clear, my intention here is not to slowly convert the actual code we write for blocks to TS. I just think it might be beneficial for us in this Library to make it easier for anyone who consumes it even in regular JS contexts)

fabiankaegy:
I will say that we are still not at the point where all of the WordPress core components ship with proper types. Some of the packages already are written in TS or have external type definitions. But there still is a way to go which will mean out types will not always be fully complete out of the box if we just extend a core component.

components/color-settings/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+	 */
+	clearable: boolean;
+	/**
+	 * Callback called when a color is selected.
+	 */
+	colors: PaletteObject[];
+	/**
+	 * Whether the palette should have a clearing button or not.
+	 */
+	onChange: (value: string) => void;
```
Antonio-Laguna: I think docs here are wrong?

components/content-picker/PickedItem.tsx
```diff
@@ -62,13 +76,16 @@ const PickedItem = ({ item, isOrderable, handleItemDelete, mode, id }) => {
 	// empty, it will return null, which is handled in the effect below.
 	const preparedItem = useSelect(
 		(select) => {
-			const { getEntityRecord, hasFinishedResolution } = select('core');
+			// @ts-ignore
+			const { getEntityRecord, hasFinishedResolution } = select(coreStore);
 
 			const getEntityRecordParameters = [type, item.type, item.id];
-			const result = getEntityRecord(...getEntityRecordParameters);
+			const result: any = getEntityRecord(...getEntityRecordParameters);
```
Antonio-Laguna, line 84: This is erroring on me `TS2556: A spread argument must either have a tuple type or be passed to a rest parameter.`

components/icon-picker/inline-icon-picker.tsx
```diff
@@ -23,16 +24,12 @@ const StyledIconPickerDropdown = styled(IconPicker)`
 export const InlineIconPicker = (props) => {
 	const { value, ...rest } = props;
 	const IconButton = useCallback(
-		({ onToggle }) => (
+		({ onToggle }: { onToggle: Function}) => (
```
Antonio-Laguna:
```suggestion
		({ onToggle }: { onToggle: Function }) => (
```

components/inner-block-slider/icons.tsx
```diff
@@ -1,3 +1,5 @@
+import React from "react";
```
Antonio-Laguna:
```suggestion
import React from 'react';
```

Antonio-Laguna (commented):
Hey Fabian!

I love this as I said in private! 

Here are some comments:

On `registerIcons`, I would update to something like: 

```ts
type Icon = {
	source: string;
	name: string;
	label: string;
};

export function registerIcons(options: Icon[]) {
	dispatch(iconStore).registerIconSet(options);
}
```

That way, whenever you try to pass something that does not belong, you would get yelled:

```
TS2345: Argument of type '{ source: string; name: string; }' is not assignable to parameter of type 'Icon[]'.   Object literal may only specify known properties, and 'source' does not exist in type 'Icon[]'.
```

I'd also add types to more places such as `PickedItem` functions so that if you use the element, you get the documentation.

I'm happy to jump here and help you with this or can also offer more direct feedback on the areas. Whatever you prefer!

This is a great addition!

fabiankaegy:
@Antonio-Laguna Thanks so much for your feedback! I addressed your immediate comments :) 

If you have the time to dive in and improve the types / add more that would be great :) But this isn't a top priority to no rush :)

fabiankaegy:
Just as an FYI, I published this branch as an alpha version here: `npm i @10up/block-components@1.15.12-alpha.0`

components/clipboard-button/index.tsx
```diff
@@ -1,17 +1,28 @@
...
+		copy?: string;
+		copied?: string;
+	};
+};
 
-export const ClipboardButton = (props) => {
+export const ClipboardButton: FC<ClipboardButtonProps> = (props) => {
 	const { text, disabled, onSuccess, labels } = props;
 	const [hasCopied, setHasCopied] = useState(false);
 	const copy = labels.copy ? labels.copy : __('Copy');
```
Manussakis:
```suggestion
	const copy = labels.copy || __('Copy');
```

components/clipboard-button/index.tsx
```diff
@@ -1,17 +1,28 @@
...
 	const copied = labels.copied ? labels.copied : __('Copied');
```
Manussakis:
```suggestion
	const copied = labels.copied || __('Copied');
```

components/counter/index.tsx
```diff
@@ -136,14 +141,20 @@ const CircularProgressBar = (props) => {
 	);
 };
 
+type CounterProps = {
+	count: number;
+	limit: number;
+	[x: string]: any;
```
Manussakis: Hey @fabiankaegy 

Is there a reason to use `x` instead of `key` here? 🤔

components/icon-picker/icon.tsx
```diff
@@ -1,27 +1,27 @@
...
+	name: string;
+	onClick?: Function | undefined;
+	iconSet: string;
+	[key: string]: any;
+};
+
+export const Icon: FC<IconProps> = (props) => {
 	const { name, iconSet, onClick, ...rest } = props;
-	const icon = useIcon(iconSet, name);
+	const icon= useIcon(iconSet, name);
```
Manussakis:
```suggestion
	const icon = useIcon(iconSet, name);
```

components/icon-picker/inline-icon-picker.tsx
```diff
@@ -1,46 +1,40 @@
...
- * @returns {*}
- */
-export const InlineIconPicker = (props) => {
+export const InlineIconPicker: FC<IconPickerProps> = (props) => {
 	const { value, ...rest } = props;
 	const IconButton = useCallback(
-		({ onToggle }) => (
-			<Icon name={value?.name} iconSet={value?.iconSet} onClick={onToggle} {...rest} />
+		({ onToggle }: RenderToggleProps) => (
+			<Icon name={value.name} iconSet={value?.iconSet} onClick={onToggle} {...rest} />
```
Manussakis: @fabiankaegy the `iconSet` property has not been defined as optional in the [`Icon` type definition](https://github.com/10up/block-components/pull/191/files#diff-944cf70fc1be5c7b4fc19291f6b09dba58a5fed031b77f7bde5061a80a8b6afeR4). You can probably remove the question mark here as you did with the `name` property.

components/image/index.tsx
```diff
@@ -1,11 +1,31 @@
...
-const Image = (props) => {
+type FocalPoint = {
+	x: number,
+	y: number,
+};
+
+type ImageProps = {
+	id: number;
+	size: string;
+	onSelect(value: { id: number } & { [k: string]: any }): void;
```
Manussakis: @fabiankaegy  I would stick with `key` instead of `x` or `k` whenever possible, just to follow a pattern. 😉

components/post-context/context.ts
```diff
@@ -0,0 +1,19 @@
+import { createContext, useContext } from '@wordpress/element';
+
+export const DEFAULT_POST_CONTEXT = {
```
Manussakis: What if you add the `PostContextType` type to the `DEFAULT_POST_CONTEXT` const to make it more type safety? 🙂

components/post-term-list/index.tsx
```diff
@@ -6,13 +6,24 @@ import {
...
 import { PostTermContext } from './context';
 import { ListItem, TermLink } from './item';
 
-export const PostTermList = (props) => {
+export type PostTermListProps = {
+	tagName?: keyof HTMLElementTagNameMap | undefined;
+	taxonomyName: string;
+	children?:
+		| React.ReactNode
+		| ((props: { selectedTerms: Array<{ id: number; name: string }>, isEditable?: boolean }) => React.ReactNode);
```
Manussakis: Here is another way to add type to the `selectedTerms ` property: `{ id: number; name: string }[]` 😉

components/post-term-list/item.tsx
```diff
@@ -1,26 +1,27 @@
 import { useContext } from '@wordpress/element';
-import PropTypes from 'prop-types';
+import { FC } from 'react';
 import { PostTermContext } from './context';
 
-export const ListItem = (props) => {
+export type ListItemProps = {
+	tagName?: keyof HTMLElementTagNameMap | undefined;
```
Manussakis: Hey @fabiankaegy Why was it necessary to add `| undefined` if you have already used the optional operator (`?`) 🤔

hooks/use-on-click-outside.ts
```diff
@@ -1,18 +1,12 @@
...
 
-/**
- * useOnClickOutside
- *
- * @param {Function} onClickOutside callback that will get invoked when the user clicks outside of the target
- * @returns {object} ref to the target element
- */
-export function useOnClickOutside(onClickOutside) {
-	const ref = useRef();
+export function useOnClickOutside(onClickOutside: (event: MouseEvent | TouchEvent) => void) {
```
Manussakis, line 3: @fabiankaegy By typing the `event` object with `React.MouseEvent<HTMLElement> | React.TouchEvent<HTMLElement>` types you can leverage [React's synthetic event system](https://react.dev/reference/react-dom/components/common#react-event-object).

fabiankaegy:
I merged the base typescript config to the main branch. So we can now create smaller incremental updates to the various components. Therefore I will close this large PR for now. 

Thank you to anyone that helped contribute to this PR. It has been incredibly helpful to figure out what to do and now we can actually apply our learnings by creating small individual PR's for the carious components

PR #193, opened 2023-01-24
fix how class names from core get passed through register block extension

fabiankaegy:
### Description of the Change

When filtering the `editor.BlockListBlock` to add a custom class name to it we need to ensure we correctly pipe through any already existing classes. This was done incorrectly. You can see an example of how core does it here: Here is an example of how it is done in core: https://github.com/WordPress/gutenberg/blob/70ba9218ae53e31d2c26689eed2e698dcea7f3aa/packages/block-editor/src/hooks/layout.js#L457-L460

### How to test the Change

All tests should pass

### Changelog Entry

> Fixed - Make sure any class names which get added via `editor.BlockListBlock` filter get passed through correctly

### Credits

Props @fabiankaegy 

### Checklist:

### DIFF
--- a/api/register-block-extension/index.js
+++ b/api/register-block-extension/index.js
@@ -3,6 +3,7 @@
 
 import { addFilter } from '@wordpress/hooks';
 import { createHigherOrderComponent } from '@wordpress/compose';
+import classnames from 'classnames';
 
 /**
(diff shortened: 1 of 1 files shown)


Antonio-Laguna (approved):
LGTM!

PR #195, opened 2023-01-24
allow usage of wildcard selector in `registerBlockExtension` allowed blocks

### LINKED ISSUE #194: Allow usage of wildcard selectors for `registerBlockExtension` allowed blocks
Currently, it is difficult to apply a block extension to all blocks. It would be great to be able to pass in a wildcard selector like `*` to make this easier.

fabiankaegy:
### Description of the Change

Allow usage of wildcard (`*`) selector for allow list of `registerBlockExtension`.

Closes #194 

### How to test the Change

Add a new block extension. Instead of passing a block name as the first param, pass `*`, and the extension should get applied to all blocks.

### Changelog Entry

> Added - Allow usage of wildcard selector in `registerBlockExtension` allow list

### Credits

Props @fabiankaegy 

### Checklist:


fabiankaegy:
I'm not sure whether we should start with just supporting the `*` wildcard or whether we should also handle things like `core/*` to select all core blocks for example. Curious whether anyone has thoughts on this :)

PR #197, opened 2023-02-09
Add support for adding inline styles with block extensions

### LINKED ISSUE #111: Add ability to add inline styles to a block via the `registerBlockExtension` API
Currently the `registerBlockExtension` API only allows you to extend blocks by adding custom class names to them. There are cases however where we need inline styles for a custom color for example. It would be nice to extend the API to accept both classnames as well as inline styles.

fabiankaegy:
### Description of the Change

This PR adds the ability for users of the `registerBlockExtension` API to also add inline styles to blocks instead of just supporting class names. 

Closes #111 

### How to test the Change

use the new `inlineStyleGenerator` function to generate inline styles that should be passed to the extended block.

### Changelog Entry

> Added - Ability to pass inline styles to `registerBlockExtension`

### Credits

Props @fabiankaegy 

### Checklist:


Antonio-Laguna (approved):
This LGTM!

My only suggestion here is to accompany this with more docs. It is not obvious what problem could this solve or how much your life could be easier with it.

While I don't think the `order` needs to be explained the same way, I think some of these functions that you added for development efficiency could get more explaining.

In any case, it can be done afterwards! This is awesome :)

PR #201, opened 2023-02-22
Improving stability across renders

### LINKED ISSUE #200: Popover in `PostTermList` component gets unintentially rerendered causing an input field to loose focus
The `PostTermList` component renders a Popover when a user clicks on the terms. In this popover there is a component that allows editors to edit the terms. 

This popover gets unintentionally rerendered when the user types any new string in the input field. This rerender causes the input field to lose focus interrupting the user flow. 

CleanShot 2023-02-22 at 20 16 23

Antonio-Laguna:
Fixes #200

Returning components from a hook is usually a bad idea. The reason is that hooks run on every render which, in this case, leads to a new `Popover` ( the WP's one ) being created. 

React tears down all its children and replaces them with new ones which cause the focus to be lost. 

So every time a component using `usePopover` renders it will tear down all the children of `<Popover>` and rebuild them because `<Popover>` is a "new" function component.

By switching to `useMemo` we're telling React to hold onto that function unless `[isVisible, popoverAnchor, ref]` changes which is what we actually want. Note that `ref` is only added to make ESLint happy, refs are stable across render and .they will never trigger a re-render

### Checklist:

### DIFF
--- a/hooks/use-popover/index.js
+++ b/hooks/use-popover/index.js
@@ -32,7 +32,7 @@ export const usePopover = () => {
 					</Popover>
 				);
 			},
-		[isVisible, popoverAnchor],
+		[isVisible, popoverAnchor, ref],
 	);
 
 	return {
--- a/hooks/use-popover/index.js
+++ b/hooks/use-popover/index.js
@@ -1,5 +1,5 @@
 import { Popover } from '@wordpress/components';
-import { useState } from '@wordpress/element';
+import { useState, useCallback, useMemo } from '@wordpress/element';
 import { useOnClickOutside } from '../use-on-click-outside';
 
 export const usePopover = () => {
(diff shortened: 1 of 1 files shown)


fabiankaegy (approved):
I was soooo close =D I had the Popover component momoized but hadn't converted the toggle function to a callback which I think invalidated the useMemo through the dependency array every time... 

This works like a charm and fixes the issue I was having :) Thank you!

hooks/use-popover/index.js
```diff
@@ -1,15 +1,15 @@
...
 export const usePopover = () => {
 	// Use internal state instead of a ref to make sure that the component
 	// re-renders when the popover's anchor updates.
 	const [popoverAnchor, setPopoverAnchor] = useState();
 	const [isVisible, setIsVisible] = useState(false);
-	const toggleVisible = () => {
-		setIsVisible(!isVisible);
-	};
+	const toggleVisible = useCallback(() => {
+		setIsVisible((visible) => !visible);
```
fabiankaegy, line 11: This here is also cool. I always forget that you can pass an updater function to the state setter and get access to the current value...

kirtangajjar: @fabiankaegy Is there a difference between `(visible) => !visible` and `!isVisible`  :thinking:

Antonio-Laguna: @kirtangajjar there's a difference!

The setter function from `useState` has two modes:

- One in which you pass the actual value you want to set.
- One in which you pass a function. In this scenario (the one I updated to)  the function gets the current value as an argument and must return the new value you want to set. By doing this we ensure no external dependencies so the callback can remain fully pure. 

I hope I clarified here!

PR #202, opened 2023-02-23
Add the ability to define a "noResultsMessage" on the PostTermList component

fabiankaegy:
### Description of the Change

Currently the `PostTermList` just renders completely empty if no terms are assigned to the post. This leads to a bad experience for the user because they are not able to user the inline control to select new terms. In order to fix this I've added a `noResultsMessage` prop to the component which has a default value.

I've also changed it so that the component only renders the no results message if the block this component is being used within also is currently selected. 

### Changelog Entry

> Added - Ability to define `noResultsMessage` on `PostTermList` component

### Credits

Props @fabiankaegy 

### Checklist:

### DIFF
--- a/components/optional/index.js
+++ b/components/optional/index.js
@@ -11,6 +11,6 @@ Optional.defaultProps = {
 };
 
 Optional.propTypes = {
-	value: PropTypes.string,
+	value: PropTypes.string || PropTypes.bool,
 	children: PropTypes.node.isRequired,
 };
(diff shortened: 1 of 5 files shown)


fabiankaegy:
@Antonio-Laguna I would love to also hear your thoughts on how this API could be made more "composable".

I had initially tried to use some kind of slot fill system with an actual `<PostTermList.NoResults>` component that you could use but found it to be overkill for this example.

But very curious whether you have any other ideas / think this is good as is :)

PR #203, opened 2023-02-23
fix ensure styles inline styles are getting applied correctly in the editor for block extensions

fabiankaegy:
### Description of the Change

Fixes how inline styles get applied in the editor when block extensions get applied that don't actually produce inline styles.

### Changelog Entry

> Fixed - How inline styles get applied in the editor when block extensions get applied that don't actually produce inline styles

### Credits

Props @fabiankaegy, @ncoetzer 

### Checklist:

- [X] All new and existing tests pass.


ncoetzer (approved):
@fabiankaegy Awesome, thank you 🚀

PR #207, opened 2023-03-03
add action to monitor WordPress dependencies

fabiankaegy:
Add a GitHub Action that tracks changes of the Dependencies generated by the DependencyExtractionWebpackPlugin


fabiankaegy:
Okay right now it looks like it doesn't build the `example/` package which is where the `.asset.php` files would be located. I can fix that next week

PR #216, opened 2023-04-05
Check if query exists before assign (React 18 useEffect issue)

s3rgiosan:
### Description of the Change

I discovered that the <ContentSearch> component is not functioning properly with WordPress 6.2, and I suspect that this is due to the shipment of React 18 (see [this](https://make.wordpress.org/core/2023/03/07/upgrading-to-react-18-and-common-pitfalls-of-concurrent-mode/), [this](https://github.com/facebook/react/issues/24502#issuecomment-1118846544), and [this](https://react.dev/learn/synchronizing-with-effects#how-to-handle-the-effect-firing-twice-in-development)). When a user types a post title quickly, the component breaks and an error is displayed in the browser's console. 

Screenshot 2023-04-05 at 22 42 32

To address this issue, I have added a check that verifies if the query exists before assigning the results.

### Changelog Entry

> Fixed - Ensure that the search query exists before assigning the results

### Credits

Props @s3rgiosan 

### Checklist:

### DIFF
--- a/components/content-search/index.js
+++ b/components/content-search/index.js
@@ -217,11 +217,17 @@ const ContentSearch = ({
 							setSearchQueries((queries) => {
 								const newQueries = { ...queries };
 
-								if (newQueries[searchQueryString]) {
-									newQueries[searchQueryString].results = normalizedResults;
-									newQueries[searchQueryString].totalPages = totalPages;
-									newQueries[searchQueryString].controller = 0;
+								if (typeof newQueries[searchQueryString] === 'undefined') {
+									newQueries[searchQueryString] = {
+										results: null,
+										controller: null,
+										totalPages: null,
+									};
 								}
+
+								newQueries[searchQueryString].results = normalizedResults;
+								newQueries[searchQueryString].totalPages = totalPages;
+								newQueries[searchQueryString].controller = 0;
 
 								return newQueries;
 							});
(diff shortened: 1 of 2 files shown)


s3rgiosan:
@fabiankaegy @Antonio-Laguna Can we get this reviewed and merged ASAP? We have at least two clients that currently are unable to use the `ContentPicker` and `ContentSearch` components because the block breaks as soon as they start typing.
This started to happen since the WP 6.2 update.

Antonio-Laguna:
@s3rgiosan thanks for digging into this and also for providing such a detailed summary!

Looking into this I'm not sure this is the best solution though I'm not sure what the solution is right now. The problem I see with this approach is that API Calls results are not saved. If they don't exist they don't get saved. 

Perhaps it would be safer to initialize the object if it didn't exist and then save it initialized and with data in it?

I don't think there's a need to handle the queries at the state level since there are 5 different places in which `searchQueries` gets updated but I think this is beyond the point at this stage

s3rgiosan:
@Antonio-Laguna thank you for your feedback. I also wasn't totally convinced that it was the best solution, but I ran out of ideas 😆  I've pushed a new update that initializes the query object if it doesn't exist. Can you check?

Antonio-Laguna:
@s3rgiosan that makes me feel better :) 

It looks good to me!

Antonio-Laguna (approved):

fabiankaegy (approved):
Also looks great to me :)

kkoppenhaver:
@fabiankaegy @s3rgiosan Thank you for this! We were running into this issue as part of a client launch and this fix saved a couple of our blocks 🎉

PR #218, opened 2023-04-16
Add ability to use drag & drop to select and replace an image in the Image component

### LINKED ISSUE #165: Add ability to drag an image onto the `Image` component to directly upload and replace the selected image
This is already how the core image block works. We should also add this behavior to the `Image` component.

Sidsector9:
### Description of the Change

Adds ability to drag and replace the image in the Image Component

Closes #165 

### Screencast

https://github.com/10up/block-components/assets/17757960/076446f5-80d3-4773-b924-97632c9a2055

### How to test the Change

1. Add an image using the Image Example block.
2. Drag a new image over the block to replace it.

### Changelog Entry

> Added - Drag to replace image in the Image component.

### Credits

### Checklist:


Sidsector9:
@fabiankaegy the functionality works without error but fails within the E2E context. I've tried debugging and got this error:

Screenshot 2023-05-29 at 2 35 58 PM

Would you have any clue about this?

fabiankaegy:
@Sidsector9 Yeah, I've also seen that issue unrelated to this PR. But I've not yet been able to figure out what the cause for that actually is :(

Sidsector9:
:( I'll spend some time this week to investigate further. I'll open this PR for review.

fabiankaegy:
@Sidsector9 Thank you for working on this :) I noticed one small regression:

 The Focal point picker now renders when no image is present, which should not be the case. 

CleanShot 2023-05-31 at 09 22 46@2x


I'm also still a little unsure how to best go about the image markup. It causes some issues that we have to introduce the additional DIV in the markup since many usages of this component rely on the image being a direct descendant of a figure element for example... We are running into the same issue here: #172 

I'm not sure we have a better solution though. Just throwing it out there.

Sidsector9:
@fabiankaegy The console error originates from WP Core only when the Block Inserter (the one on the top left of the editor) is used. This happens as the patterns start to load. I did not dig into Core to investigate further, I instead used an alternative way to resolve it. The solution is related to https://github.com/10up/cypress-wp-utils/issues/96

See the video below:


https://github.com/10up/block-components/assets/17757960/a3c9ebc5-6c15-46b6-8dc5-d9331c232918

--

As for the regression with the focal point picker, that is resolved.

> I'm also still a little unsure how to best go about the image markup. It causes some issues that we have to introduce the additional DIV in the markup since many usages of this component rely on the image being a direct descendant of a figure element

That makes sense. On a high level #222 describes good solutions, I'll dig further and respond if I can think of any more solutions.

fabiankaegy:
Closing this as stale for now

PR #219, opened 2023-04-19
Use switch/case in Link

szepeviktor:
This is code improvement "only".

1. This change is so obvious
2. It was so much joy to send this PR

### Changelog Entry

Changed - Use switch/case in Link

### Checklist:

### DIFF
--- a/components/link/index.js
+++ b/components/link/index.js
@@ -38,13 +38,14 @@ function getSuggestionsQuery(type, kind) {
 		case 'post_format':
 			return { type: 'post-format' };
 		default:
-			if (kind === 'taxonomy') {
-				return { type: 'term', subtype: type };
+			switch (kind) {
+				case 'taxonomy':
+					return { type: 'term', subtype: type };
+				case 'post-type':
+					return { type: 'post', subtype: type };
+				default:
+					return {};
 			}
-			if (kind === 'post-type') {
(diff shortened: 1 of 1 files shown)


fabiankaegy:
On general I would agree. However nested switch statements do get somewhat confusing. So closing this for now

szepeviktor:
Yes. But I do not support those users/developers for whom it is somewhat confusing.

szepeviktor:
> I do not support

Meaning I do not allow them to influence my work.
Otherwise I would have to write 👶🏻 baby code.
But I have to admit: in the WordPress ecosystem you don't really meet software engineers.
The WordPress ecosystem is frustrating for engineers as you are unable to fly, just flap your wings and run as fast as you can.

fabiankaegy:
Hi @szepeviktor 👋 

I want to remind you that this project has a [Code of Conduct](https://github.com/10up/block-components/blob/develop/CODE_OF_CONDUCT.md). I have hidden your previous replies because they are not following those rules. 

If you have constructive criticism/reasons why you think my decision to close your PR here you are more than welcome to share those to try and change my mind.

Thank you in advance!

szepeviktor:
> I have hidden your previous replies

You are right Fabian. Telling those facts was really abusive.

PR #221, opened 2023-05-24
Clean up content-search component and re-search when content types change

### LINKED ISSUE #213: ContentPicker / ContentSearch - changing content type does not trigger a new search
### Describe the bug

The ContentSearch component will run a new search when the search term changes. But if the search term stays the same while the content type changes, a new search is not run; instead the spinner shows indefinitely. Below is a minimal test case to show the issue.

**Expected behavior:**  When the content type changes, the a new search query should run.

I'd be happy to help but probably not for a couple weeks, so no problem if someone else wants to pick it up.

Side note, I have looked over the ContentSearch component and feel that it needs some general cleanup beyond this bug (modernizing syntax, rearchitecting to eliminate the need for ESLint ignores, particularly with exhaustive dependencies), but that's probably a separate issue.

### Steps to Reproduce

1. Add the block code snippet to your WP theme 
2. Insert the block in the editor
3. Enter a search term
4. Change the post type
5. See the infinite spinner

```JS

import { registerBlockType } from '@wordpress/blocks';
import { useState } from '@wordpress/element';
import { ContentPicker } from '@10up/block-components';

const Edit = () => {
	const [posts, setPosts] = useState([]);
	const [postType, setPostType] = useState('post');

	return (
		<>
			<select onChange={(e) => setPostType(e.target.value)} value={postType}>
				<option value="post">Post</option>
				<option value="page">Page</option>
			</select>
			<ContentPicker
				content={posts}
				label="Search Posts"
				onPickChange={setPosts}
				contentTypes={[postType]}
			/>
		</>
	);
};

// Register the test block type.
registerBlockType('jw/test', {
	title: 'ContentPicker Test',
	icon: 'admin-post',
	category: 'widgets',
	edit: Edit,
	save: () => null,
});
```

### Environment information

_No response_

### WordPress information

WP 6.2

### Code of Conduct

johnwatkins0:
Fixes #213 

### Description of the Change

This makes several changes that will improve the performance of the ContentSearch component and make it behave more predictably.  It also modifies the component to account for changes to the contentTypes prop. I think this should just be phase 1 because the component is still too large, and the use of useRef to detect changes to the contentTypes prop is not ideal (component should be rearchitected to handle changes to the prop with effect hooks).

I recommend covering this with tests. I made an attempt but could tell it was going to be a rabbit-hole so didn't stick with it 😄   With the size of this change I would understand holding this until I or someone else can write those tests.

Changes:

1. Reduce unnecessary renders by (a) moving variables not reliant upon state or props up to the top level of the module, and (b) move some functions to the top level of the module (what were previously useCallback dependencies now passed in as arguments).
2. Put the mounted ref to use (it was previously never used)
3. Added `fetchedInitialResults` state to eliminate the need to ignore the exhaustive deps ESLint rule
4. Replaced .then Promise syntax with async/await
5. Wrap some complex logic that was running on every render in a useMemo

### How to test the Change

See the test block I shared in #213. A new query should run when you change the content type.

Additionally, some regression testing should be performed with this branch against existing features using the ContentSearch omponent.

### Changelog Entry

Clean up ContentSearch component and handle changes to the contentType prop

### Credits

Props @johnwatkins0

### Checklist:

### DIFF
--- a/components/content-search/index.js
+++ b/components/content-search/index.js
@@ -1,6 +1,6 @@
 import { TextControl, Spinner, NavigableMenu, Button } from '@wordpress/components';
 import apiFetch from '@wordpress/api-fetch';
-import { useState, useRef, useEffect, useCallback } from '@wordpress/element';
+import { useState, useRef, useEffect, useCallback, useMemo } from '@wordpress/element';
 import PropTypes from 'prop-types';
 import { __ } from '@wordpress/i18n';
 // eslint-disable-next-line no-unused-vars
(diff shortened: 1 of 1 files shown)


fabiankaegy:
Closing in favor of #326

PR #231, opened 2023-06-16
Search content when at least 3 letters typed in ContentSearch component

vishalkakadiya:
### Description of the Change
Adds functionality to search the content when at least three letters are typed in the ContentSearch's component search functionality, so we can save network bandwidth.

### How to test the Change
- Add the "Post Searcher" component to the page.
- Type just one letter the REST API is firing, so it is good practice to fire a REST request when at least three letters are added in the search.

### Credits
Props @vishalkakadiya

### DIFF
--- a/components/content-search/index.js
+++ b/components/content-search/index.js
@@ -14,17 +14,17 @@ const NAMESPACE = 'tenup-content-search';
 const listMinHeight = '46px';
 
 const ContentSearch = ({
-	onSelectItem,
-	placeholder,
-	label,
-	contentTypes,
-	mode,
-	perPage,
-	queryFilter,
-	excludeItems,
-	renderItemType,
-	fetchInitialResults,
-}) => {
+						   onSelectItem,
+						   placeholder,
+						   label,
+						   contentTypes,
+						   mode,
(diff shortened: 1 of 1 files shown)


Antonio-Laguna:
Hey @vishalkakadiya thanks for raising this!

While I sort of agree with your point, this would be a breaking change since usage could be relying on this. IMHO in the admin scenario in which lots of things are requesting data without you even knowing it's not the end of the world for this to happen.

I think as it stands we wouldn't go with this change. One way I can see to make this potentially work is by adding a prop. Something like: `minQueryChars` which by default is `0` so it would match but you'd have control to make that number `3` or whatever the number makes sense for your site.

Thoughts?

vishalkakadiya:
@Antonio-Laguna Thank you for quickly checking on this. 🙌 

Yes, it totally makes sense to build this as the props, let me know if you want me to update this PR to work with that prop.

Thanks! 🙂

Antonio-Laguna:
@vishalkakadiya please go ahead! Thanks for be willing to tackle this!

fabiankaegy:
Closing this as stale for now

PR #235, opened 2023-06-29
feat/234: Add `Required` Component

### LINKED ISSUE #234: Create a new `Required` component
It would be great to have a simple react component mark any component as required. It should work similar to the `<Optional>` component we already have in that you pass it a value that needs to exist in order for the condition to be met. 

If the condition is not met, but the block is in the document, the editor should be locked from saving anything using the core APIs. 

```js
const { lockPostSaving, unlockPostSaving } = useDispatch( 'core/editor' );

// we can then lock / unlock the editor using these functions here
lockPostSaving( 'title-empty-lock' );
unlockPostSaving( 'title-empty-lock' );
```

Removing the block should also unlock the editor. 

Additionally, it would be great to show the reason why the editor is locked in a notice when the user tries to save the post & also something in the `PrePublishCheck` to show all the required fields that are empty would be great. 
```[tasklist]
### Tasks
- [ ] Lock the editor when a block containing at least one `Required` component is rendered
- [ ] Unlock the editor when all `Required` components in the entire editor have their conditional met (Each item will handle it's own locked state)
- [ ] Unlock the editor when the block containing the required field gets removed.
```

I think we should be able to use the clean up function of a `useEffect` hook to handle the unlocking on removal for us.

```js
useEffect( () => {

    if (!condition) {
        // lock the editor
    } else {
        // unlock the editor
    }

    return () => {
        // unlock the editor
    }

}, [condition] )
```

Sidsector9:
### Video

https://github.com/10up/block-components/assets/17757960/bdf770dd-bbd1-43fb-9295-cc224e058f0e

Closes #234 

### Changelog Entry

> Added - `Required` component

### Credits

### Checklist:


Sidsector9:
@fabiankaegy I took a slightly different approach to show the error messages. (Video added in the PR description)

I created a custom data store that keeps track of all the Blocks that uses the `Required` component and adds an inline error message under the block if the required field(s) are not set.

Wanted to check with you if you have a better approach before I continue with writing the tests.

fabiankaegy:
Hey @Sidsector9 👋 

This is super cool :) 

Can you share a little more detail about why you decided to go with this approach? Was it because the other approach didn't work or because the messaging for users was unclear/confusing? 

Would love to better understand the constraints here :)

Sidsector9:
Sure @fabiankaegy. 

> Was it because the other approach didn't work?

I was able to lock/unlock the post by providing unique IDs to a `Required` component's `lockPostSaving()` and `unlockPostSaving()` functions.
So even if a post contains multiple blocks, each block with multiple `Required` components, the post will remain locked until all `Required` components trigger `unlockPostSaving()` with their unique ID. It worked so far.

> Why you decided to go with this approach?

1. I couldn't find a way to send the information about the block containing the `Required` component to the Pre publish panel for displaying the notice without a custom data store.
2. The Pre publishing panel is only available for a new unpublished post. Once it is published, the "Update" button is disabled out. I thought of adding the notices to the PostStatus panel but didn't go through with it to avoid clutter.
3. I thought it would be cool to highlight the block as soon as it fails validation, and not wait for the editor to reach the Prepublish panel. This way, we avoid the implementation of showing the warning/error in the Prepublish panel and Notices.
4. This also works well when the input fields of a Block are hidden by default and only rendered when the Block is selected.

I would love to hear if you have a better approach, somewhere I strongly believe you do 😄

fabiankaegy:
@Sidsector9 Thank you again for working on this! It does exactly what I was looking and hoping for. I'm going to close this PR anyways for now, as @darrenjacoby came up with a really cool more powerful way to solve the same issue in #238 

I still like the cleanliness of the `Required` component. But having the ability for custom validations is really powerful and at the same time also solves all the things we were tying to create here.

Sidsector9:
I just saw @darrenjacoby's PR and I agree, it does a lot of other validations that this PR doesn't. Completely on board with the decision 👍

PR #238, opened 2023-07-07
New component: FieldValidation

### LINKED ISSUE #234: Create a new `Required` component

darrenjacoby:
### Description

This PR creates a new `<FieldValidation>` component that can be used for required fields or fields that require custom tests for validation.

- Fields can be marked as required.
- Fields can have custom validation tests and responses.
- Editor "update" button is disabled if any field validation is failing on the post or page.

Screenshare with failing field validations;

https://github.com/10up/block-components/assets/15921694/18e43322-2236-46a6-abbd-fad0280a48cc

### Testing

Import `<FieldValidation>` and test with the following;

```js
import { FieldValidation } from '@10up/block-components';

function BlockEdit(props) {
    const { title } = attributes;

    return (
        <>
            <FieldValidation
                value={title}
                required={__('Title is a required field', '10up')}
                validate={[
                    [ (value) => /^[a-zA-Z\s]+$/.test(value), __('Title requires alphabetical characters', '10up') ],
                    [ (value) => /^.{10,}$/.test(value), __('Title requires 10 characters or more', '10up') ],
                ]}
            >
                    <RichText
                        tagName="p"
                        value={title}
                        onChange={(value) => setAttributes({ title: value })}
                    />
            </FieldValidation>
        </>
    )
}
```

**Known issue:** `<RichTextCharacterLimit>` also uses floating-ui and seems to impact the ref used by `<FieldValidation>` when nested, impacting the position of validation responses. Solution requires further investigation.

### Changelog
- Added - Field Validation component

### Credits
n/a

### Checklist

### DIFF
--- a/components/field-validation/index.js
+++ b/components/field-validation/index.js
@@ -22,7 +22,8 @@ const createTestsArray = (required, validate) => {
 
 	if (required) {
 		const fn = (str) => typeof str === 'string' && str !== '';
-		const res = typeof required === 'string' ? required : __('Required', '10up-block-components');
+		const res =
+			typeof required === 'string' ? required : __('Required', '10up-block-components');
 		tests.push([fn, res]);
 	}
 
@@ -88,9 +89,16 @@ const Error = forwardRef((props, ref) => {
 	const { responses } = props;
 
 	return (
-		<ErrorResponse className="tenup--block-components__validation-error__response" {...props} ref={ref}>
+		<ErrorResponse
+			className="tenup--block-components__validation-error__response"
+			{...props}
+			ref={ref}
+		>
 			{responses.map((response) => (
-				<div key={uuid()} className="tenup--block-components__validation-error__response-rule">
+				<div
+					key={uuid()}
+					className="tenup--block-components__validation-error__response-rule"
+				>
 					{response}
 				</div>
 			))}
@@ -156,14 +164,15 @@ const FieldValidation = (props) => {
 		dispatchLock(responses.length > 0);
 	}, [responses, dispatchLock]);
 
-	const fieldErorrClassName = responses.length > 0 ? 'tenup--block-components__validation-error__field' : '';
+	const fieldErrorClassName =
+		responses.length > 0 ? 'tenup--block-components__validation-error__field' : '';
 
 	return (
 		<>
 			{cloneElement(children, {
 				ref: reference,
 				...children.props,
-				className: classnames(children.props?.className, fieldErorrClassName),
+				className: classnames(children.props?.className, fieldErrorClassName),
 			})}
 
 			{responses.length > 0 && (
--- a/components/field-validation/index.js
+++ b/components/field-validation/index.js
@@ -1,0 +1,189 @@
+import { __ } from '@wordpress/i18n';
+import { useState, useEffect, useMemo, forwardRef, cloneElement } from '@wordpress/element';
+import { dispatch } from '@wordpress/data';
+import { useFloating, autoUpdate } from '@floating-ui/react-dom';
+import { v4 as uuid } from 'uuid';
+import PropTypes from 'prop-types';
+import styled from '@emotion/styled';
+
+/**
+ * Create Tests Array
+ *
+ * @description create a sanitized tests array.
+ *
+ * @param {boolean|string} required Field validation required prop.
+ * @param {boolean|Array} validate Field validation validate prop.
+ *
+ * @returns {Array}
+ */
+const createTestsArray = (required, validate) => {
+	const tests = [];
+
+	if (required) {
+		const fn = (str) => typeof str === 'string' && str !== '';
+		const res = typeof required === 'string' ? required : __('Required', '10up-block-library');
+		tests.push([fn, res]);
+	}
+
+	if (validate && Array.isArray(validate)) {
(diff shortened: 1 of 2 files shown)


components/field-validation/index.js
```diff
@@ -0,0 +1,189 @@
...
+ * @param {boolean|Array} validate Field validation validate prop.
+ *
+ * @returns {Array}
+ */
+const createTestsArray = (required, validate) => {
+	const tests = [];
+
+	if (required) {
+		const fn = (str) => typeof str === 'string' && str !== '';
+		const res = typeof required === 'string' ? required : __('Required', '10up-block-library');
```
fabiankaegy:
```suggestion
		const res = typeof required === 'string' ? required : __('Required', '10up-block-components');
```

fabiankaegy (changes requested):
This is really cool and works like a charm! Thank you for your contribution! 

The only thing I would like to workshop a little bit more is how we style the inline errors. Using floatingUI is great and solves most of our issues already. However I think we should maybe think about making the entire background a solid color so we can ensure legibility. 

@Sidsector9 implemented a similar thing in https://github.com/10up/block-components/pull/235 which we may be able to use as an inspiration. 

Maybe @xavortm also has an idea of what this could look like :)

components/field-validation/index.js
darrenjacoby: Oops, corrected.

darrenjacoby:
Hey @fabiankaegy ,

Thanks for taking a look! There's probably a good case to be made for leaving styling options to developers on each project to account for things like dark-mode/brand colors, etc. But definitely open to design suggestions for defaults.

In the interim, pushed an update which implements an error class on the field itself.

- `.tenup--block-components__validation-error__field` to style the field. For example, adding `background-color: red;` will made the field background red.

image

There are two other available custom classes;

- `.tenup--block-components__validation-error__response` to style the response container.
- `.tenup--block-components__validation-error__response-rule` to style each individual response rule (as there can be multiple rule failures).

fabiankaegy:
I love the addition of the classes. 

However, I kind of disagree that it should be up to the developer to determine how the invalid state should look. The validation issue is not part of the content and counts as part of the editor frame. Having a standardized way to showcase invalid content which then can be overwritten, but in most cases doesn't need to be overwritten, is what I would prefer here. 

The goal of the block components is to remove as much complexity from the developers on individual projects. And having to re-invent how to style the invalid state of a component in the editor doesn't add a lot of value to a client in my perspective. And it creates the risk that when not styled correctly we create very unclear confusing experiences where a user doesn't understand why they are not able to save the post. 

That isn't to say that I don't want anyone to be able to overwrite the default styling. But I think there should be decent opinionated default styling that makes it look as though it were a core API and part of the native editor experience. So using the editors font, color scheme etc.

darrenjacoby:
@fabiankaegy those are good points which all make sense to me! Let’s see how we can implement a nice default state 🙂

darrenjacoby:
I've implemented some basic styling which matches some WordPress defaults. I'll await feedback from [xavortm](https://github.com/xavortm).

Responses;
- Default WordPress font stack
- Font size 13px
- Font weight 500
- WordPress color `var(--wp--preset--color--vivid-red)`

Field;
- Background color of `var(--wp--preset--color--vivid-red)` at 75% opacity.

Example on dark background; 
image

Example on light background;
image

xavortm:
Hi! Joining with some design proposals about the validation UI/UX:

Please see here: https://www.figma.com/file/YtfzsSMXZJb2N2pfnuuAI3/Blocks-validation?type=design&node-id=0%3A1&mode=design&t=HeaSOLhTcoyL21Rc-1

I have listed the steps and states, including singular and combinations blocks (like core/blocks). 

For visual guide/direction, please see the blue section marked as "Ready for dev" on the right. Here is also a screenshot from it:

Screenshot 2023-08-03 at 10 30 20@2x

It uses a component from Gutenberg but also adds a slight shadow to work better on white backgrounds. The Figma docs point to https://developer.wordpress.org/block-editor/how-to-guides/notices/ for the component reference.

PR #243, opened 2023-08-04
Feature | Improve Icon Picker by virtualizing icon grid and using fuzzy search

fabiankaegy:
### Description of the Change

Improve the quality of search results in Icon Picker by switching to using a fuzzy search algorithm used inside `useFilteredList`
Also updates Icon Picker grid of icons to using virtualized rendering of Icons to improve the performance of large icon sets.

Closes #

### How to test the Change

Use the `Icon Picker Example` component from the example blocks

### Changelog Entry

> Fixed - Search use fuzzy search algorithm used inside `useFilteredList` to improve quality of search results
> Fixed - Use virtualized rendering of Icons in IconPicker Grid to improve the performance of large icon sets

### Credits

Props @fabiankaegy 

### Checklist:

### DIFF
--- a/components/icon-picker/icon-picker.js
+++ b/components/icon-picker/icon-picker.js
@@ -5,8 +5,8 @@ import {
 	CheckboxControl,
 	BaseControl,
 	NavigableMenu,
-	VisuallyHidden,
 	SearchControl,
+	Tooltip,
 } from '@wordpress/components';
 import { useInstanceId } from '@wordpress/compose';
 import { useState, memo, useMemo } from '@wordpress/element';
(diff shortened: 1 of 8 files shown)


Antonio-Laguna (approved):
This LGTM! 

`uFuzzy` looks very useful in this scenario, well thought!

components/icon-picker/icon-picker.js
```diff
@@ -121,15 +113,16 @@ IconPicker.propTypes = {
...
-		</>
+		<Tooltip text={icon.label}>
+			<div>
+				<StyledIconButton
+					selected={isChecked}
+					key={icon.name}
+					name={icon.name}
+					iconSet={icon.iconSet}
+				/>
+			</div>
```
fabiankaegy: The reason I add a `div` here is that the `ToolTip` component tries to clone the element passed in in order to add a new ref to it. This fails on the styled components element because it doesn't forward the ref. 

So essentially if there is a better solution for passing one single element here that forwards the ref I'm all for it :)

PR #245, opened 2023-08-09
fix add context view to rest request so that authors can view terms

fabiankaegy:
### Description of the Change

Whilst testing the PostTermList component on a site we found an issue where the block crashed for anyone with the Author user role. Traking it down, the issue is that authors are not able to access the edit context of rest API requests to the taxonomy endpoint. To fix it we explicitly set the context in the `useAllTerms` hook to `view`

### How to test the Change

1. Create a user with the "Author" role
2. Create a new post as the Author
3. Insert the "Hero" block bundled with the example plugin
4. See that the block fails. Test again with this patch applied and see that it is ressolved

### Changelog Entry

> Fixed - Change context of rest query in `useAllTerms` hook to `view` to ensure users with the role Author can use the component.

### Credits

Props @dborgesmj @RobertWise  @fabiankaegy 

### Checklist:

### DIFF
--- a/hooks/use-all-terms/index.js
+++ b/hooks/use-all-terms/index.js
@@ -11,6 +11,7 @@ export const useAllTerms = (taxonomyName) => {
 				taxonomyName,
 				{
 					per_page: -1,
+					context: 'view',
 				},
 			];
 


Antonio-Laguna (approved):
LGTM, nice catch!

PR #249, opened 2023-08-23
Fix Icon Picker rendering empty when one item should be rendered

### LINKED ISSUE #248: Icon Picker renders empty when only one item should be rendered
### Describe the bug

When using the IconPicker component in 1.16.1, icons that are registered via `registerIcons` method are not rendering in search results.
image
image

### Steps to Reproduce

1. Create a block utilizing the <IconPicker> component
2. Register some custom icons using the code below
```import { registerIcons } from '@10up/block-components';

registerIcons({
    name: 'hgf/icons',
    label: __('HGF Icons', 'hgf'),
    icons: [
        {
            source: '<svg height="50" width="50"><circle cx="50" cy="50" r="40" stroke="black" stroke-width="3" fill="red" /></svg>',
            name: 'twitter',
            label: 'Twitter',
        }
    ],
});
```
4. Switch from `1.15.13` to `1.16.1`
5. Note icon renders in `1.15.13` but not `1.16.1`

_No response_

### Environment information

_No response_

### WordPress information

6.3

### Code of Conduct

fabiankaegy:
### Description of the Change

The code to retrieve the icon to be rendered in the icon grid had a logical error in it. When it tried to calculate the index of the icon to render based on the position of the icon in the virtualized grid, the maths was adding +1 at the end to make its index human-readable instead of index 0 which we would need to actually render the correct icon. 

This issue means that regardless of the state of the icon picker, the first icon in the list is not rendered. This becomes most obvious when there only is one icon like when you are searching and there is only one match.

Closes #248 

### How to test the Change

1. Insert Icon Picker
2. Search for an icon name that only occurs once. / Alternatively only register one singular icon

### Changelog Entry

> Fixed - Empty rendering of first item in Icon Picker due to index 0

### Credits

Props @psorensen @TylerB24890 @fabiankaegy 

### Checklist:

### DIFF
--- a/components/icon-picker/icon-picker.js
+++ b/components/icon-picker/icon-picker.js
@@ -153,7 +153,7 @@ IconLabel.propTypes = {
 const IconGridItem = memo((props) => {
 	const { columnIndex, rowIndex, style, data } = props;
 	const { icons, selectedIcon, onChange } = data;
-	const index = rowIndex * 5 + columnIndex + 1;
+	const index = rowIndex * 5 + columnIndex;
 	const icon = icons[index];
 	const isChecked = selectedIcon?.name === icon?.name && selectedIcon?.iconSet === icon?.iconSet;
 


fabiankaegy:
I have pushed this fix to NPM as an alpha version so we can all test it on actual projects. To pull down the specific version run

```bash
npm i @10up/block-components@1.16.2-icon-picker.0
```

Antonio-Laguna (approved):

Antonio-Laguna:
@fabiankaegy this LGTM!

TylerB24890 (approved):
Confirmed this seems to be working for me!

PR #257, opened 2023-08-28
Fix JSX nesting error in example

markhowellsmead:
The nesting in the JSX in the README example is invalid. The TextControl and Button components need to be wrapped in a containing element.


fabiankaegy (approved):
Great catch! Thank you for the update :)

PR #262, opened 2023-09-19
fix styling of content picker component inside iframed editors

fabiankaegy:
### Description of the Change

The `ContentPicker` & `ContentSearch` components both use emotion styled-components for styling in the editor. Emotion works by dynamically generating custom class names and style tags containing the styles for these classes. 

When the block editors get rendered in the iframed editor the style tags get injected in the main window and therefore have no effect on the content within the iframe. 

The Emotion docs mention that we need to create a new `emotion/cache` and pass it the node which it should reference for injecting the style tags.

Since this will be needed for all our components eventually, I have created a new wrapper component called `StyledComponentContext`, which is private to the block components library (it doesn't get exported). This wrapper handles the creation of this cache and ensures that the component is wrapped in the appropriate provider. 

| Before | After |
|--------|------|
| CleanShot 2023-09-19 at 10 09 45@2x | CleanShot 2023-09-19 at 10 09 08@2x |

> **Note**
> This solution is a little hacky because in order to access the actual DOM where the style tags should get injected, we need to add a ref somewhere in the same context as the component we want to render. The way this is done here is by rendering an additional `span` element next to the actual component wrapped in the provider that is set to `display: none;`
> If anyone has a better idea for how achieve this I'm also happy with that :) 

### How to test the Change

- Open the site editor
- Insert the "Hello World" Sample block
- See whether the component renders as expected

### Changelog Entry

> Fixed - Rendering of Content Picker / Content Search component in iframed editors

### Credits

Props @fabiankaegy @tlovett1 

### Checklist:

### DIFF
--- a/components/content-picker/index.js
+++ b/components/content-picker/index.js
@@ -6,6 +6,7 @@ import { __ } from '@wordpress/i18n';
 import { v4 as uuidv4 } from 'uuid';
 import { ContentSearch } from '../content-search';
 import SortableList from './SortableList';
+import { StyledComponentContext } from '../styled-components-context';
 
 const NAMESPACE = 'tenup-content-picker';
 
@@ -13,7 +14,7 @@ const NAMESPACE = 'tenup-content-picker';
  * Unfortunately, we had to use !important because on PickedItem we couldn't @emotion/styled css
  * as it was breaking sortability from react-sortable-hoc
  */
-const StyleWrapper = styled('div')`
+const StyleWrapper = styled.div`
 	& .block-editor-link-control__search-item {
 		cursor: default;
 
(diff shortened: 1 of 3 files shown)


fabiankaegy:
@Antonio-Laguna Would love to get your thoughts on this :) 

I'm somewhat worried by the [size increase](https://github.com/10up/block-components/pull/262#issuecomment-1725034226) but I don't currently see any way around it 🤔

Antonio-Laguna (approved):
This LGTM.

IMHO size doesn't matter too much in an admin screen. That said the only way out would be to move out of Styled Components but I feel that's out of the question for now.

Clever trick with the span!

PR #263, opened 2023-09-19
Fix: Make `useIsSupportedMetaField` hook context aware

fabiankaegy:
### Description of the Change

We currently check whether the currently edited post supports a meta key before actually rendering the content of the post meta component. This however means that it actually is incorrect for any post meta block used within a query or post picker block. 

This PR fixes that and ensures that the check always uses the post based on the available post context.

### How to test the Change

 - Place the Post Meta block inside a query block and see that is pulls in the correct info

### Changelog Entry

> Fixed - Make `useIsSupportedMetaField` hook context aware

### Credits

Props @fabiankaegy 

### Checklist:

### DIFF
--- a/example/src/blocks/post-meta/block.json
+++ b/example/src/blocks/post-meta/block.json
@@ -9,5 +9,9 @@
 			"default": ""
 		}
 	},
+	"usesContext": [
+		"postId",
+		"postType"
+	],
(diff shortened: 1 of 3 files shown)


Antonio-Laguna (approved):
LGTM!

PR #269, opened 2023-10-06
Add inline media controls for the Image component

ncoetzer:
### Description of the Change

This pull request adds all the functionality from https://github.com/10up/block-components/pull/172 into https://github.com/10up/block-components/pull/222

Closes https://github.com/10up/block-components/pull/172

### How to test the Change

Please refer to https://github.com/10up/block-components/pull/172

### Changelog Entry

> Added - Ability to show inline media controls for the Image component

### Credits

Props @fabiankaegy 

### Checklist:


components/image/child-components/figure.js
```diff
@@ -0,0 +1,25 @@
+import PropTypes from 'prop-types';
+import { StyledComponentContext } from '../../styled-components-context';
+import { InlineControlsStyleWrapper } from '../styles';
+
+export const Figure = (props) => {
+	const { style, children, ...rest } = props;
+
+	return (
+		<StyledComponentContext cacheKey="tenup-component-image">
+			<InlineControlsStyleWrapper {...style} {...rest}>
```
fabiankaegy: Would this also work as a singling if the image tag? 

I'd love to not add an additional div around the image :)

ncoetzer: @fabiankaegy Think this is where I got confused and am not sure I'm following 😆. Apologies! So, the `Figure` component already had the `figure` wrapper before the change:
```
const Figure = (props) => {
	const { children, style, ...rest } = props;

	return (
		<figure style={{ position: 'relative', ...style }} {...rest}>
			{children}
		</figure>
	);
};
```
I basically just swapped it for a styled component that also renders a figure.

Should there not be other conditional log in the main [component](https://github.com/10up/block-components/pull/269/files#diff-014d8ab1b491f70d246d8f0179a70772b6d2645a576e1e9859ac6c2c49a70783R104-R118)? Or am I missing something completely.

fabiankaegy: @ncoetzer Okay I may have misunderstood the code then. 

What I was trying to get at is that the end markup should look like this: 

```html
<figure>
    <img />
    <div>
        <!-- Insert anything else here -->
    </div>
</figure>
````

and not:

```html
<figure>
    <div>
         <img />
        <!-- Insert anything else here -->
    </div>
</figure>
```

ncoetzer: @fabiankaegy Aah okay cool. In that case, yes, that is how it currently works and outputs the code. 
image

However, for any existing instances of this component, it would also wrap their `` tags inside a `<figure>`, which is a breaking change if I understood things correctly. So, I'm thinking of changing the logic in the [main component](https://github.com/10up/block-components/pull/269/files#diff-014d8ab1b491f70d246d8f0179a70772b6d2645a576e1e9859ac6c2c49a70783R107-R119) to conditionally check if inline controls are needed, and if so, wrap the `` in a `<figure>`. Otherwise, it will just output the `` directly as it currently does. We could do this with any other feature placed behind a flag as well.

Of course, this new API also accommodates creating custom `children`, so the user would be able to override anything we have with anything they need.

Thoughts?

fabiankaegy (approved):
This is such a nice improvement!

PR #280, opened 2024-01-17
Replace `TextControl` with `SearchControl` in `ContentSearch` Component

psorensen:
### How to test the Change

1. Insert Content Picker or Post Picker block in /example 
2. Confirm search component is used and block is operating normally

### Changelog Entry

- Replaces TextControl with SearchControl in 'content-search' component
- Adds Content Picker block to test and demonstrate content-picker component
- Updates documentation to include new prop and updates demo gif

### Credits

Props @psorensen 

### Checklist:


Antonio-Laguna (approved):
This LGTM! Thanks @psorensen !

Will let @fabiankaegy chime in since he might more opinion here!

components/content-search/index.js
```diff
@@ -343,12 +344,13 @@ const ContentSearch = ({
 	return (
 		<StyledComponentContext cacheKey="tenup-component-content-search">
 			<NavigableMenu onNavigate={handleOnNavigate} orientation="vertical">
-				<TextControl
-					label={label}
+				<SearchControl
```
fabiankaegy, line 347:
```suggestion
				<SearchControl
					__next40pxDefaultSize
```

In order to look closer to all the core component let also use the future default style now already :)

components/content-search/index.js
```diff
@@ -357,6 +359,8 @@ const ContentSearch = ({
 					onBlur={() => {
 						setIsFocused(false);
 					}}
+					size="compact"
```
fabiankaegy: Actually I think for us here the not compact size is better

components/content-search/index.js
```diff
@@ -357,6 +359,8 @@ const ContentSearch = ({
...
+					style={{ border: '1px solid rgb(30, 30, 30)' }}
```
fabiankaegy: I'm not sure we need this manual border. With the greater space it has this should look better without I think

fabiankaegy (changes requested):
Thanks for making this update :) 

I've left some suggestions for what I personally think looks better and feels more native in the editor.

| Before | After |
|--------|------|
| CleanShot 2024-01-18 at 11 44 54@2x | CleanShot 2024-01-18 at 11 44 26@2x |

components/content-search/index.js
psorensen: Removed

components/content-search/index.js
psorensen: My concern here is that I cannot see the border of the input at all - maybe it's my eyes, but I don't even realize it's a search input without the border

psorensen: Ah - there was something with my monitor - probably a high contrast setting (ironic enough) that was preventing me from seeing the gray background of the input. I've removed the manual style.

fabiankaegy (approved):

PR #297, opened 2024-03-01
upkeep: add TS support for `useIsPluginActive` hook

Sidsector9:
### Description of the Change

Converts the useIsPluginActive hook to TypeScript.

Closes #

### Changelog Entry

> Changed - Add TypeScript support to `useIsPluginActive`

### Credits

### Checklist:


hooks/use-is-plugin-active/index.ts
```diff
@@ -0,0 +1,29 @@
+import { useSelect } from '@wordpress/data';
+import { store as coreStore } from '@wordpress/core-data';
+import type { Plugin } from '@wordpress/core-data';
+
+const ACTIVE_STATUSES: string[] = ['active', 'network-active'];
```
fabiankaegy: I think I would add as const here to get even better types :)

hooks/use-is-plugin-active/index.ts
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @param pluginName The name of the plugin to check.
+ * @returns A tuple containing two boolean values: the first indicating whether the plugin is active, 
+ * and the second indicating whether the resolution for the plugin has finished.
+ */
+export const useIsPluginActive = (pluginName: string): [boolean, boolean] => {
+	return useSelect(
+		(select) => {
+			const storeSelectors = select(coreStore);
+			const plugin: Plugin = (storeSelectors as any).getPlugin(pluginName);
```
Sidsector9, line 18: @fabiankaegy I've used `any` here as `getPlugin` and `hasFinishedResolution` are convenience methods which are dynamically named.

I tried using a custom interface:

```tsx
interface CoreStore {
	getPlugin: (pluginName: string) => Plugin;
	hasFinishedResolution: (storeName: string, resolution: any[]) => boolean;
}
```

but I am not sure which type to extend from so that `CoreStore` contains all selectors from `select(coreStore)`.

hooks/use-is-plugin-active/index.ts
```diff
@@ -0,0 +1,29 @@
...
+			const hasResolvedPlugins: boolean = (storeSelectors as any).hasFinishedResolution('getPlugin', [
+				pluginName,
+			]);
+
+			const isPluginActive: boolean = ACTIVE_STATUSES.includes(plugin?.status);
+
+			return [isPluginActive, hasResolvedPlugins] as [boolean, boolean];
```
fabiankaegy: Why is the as required here? Shouldn't it already know that due yo the earlier types?

Sidsector9: That's what confused me as well. Without it TypeScript errored saying:

> Type 'boolean[]' is not assignable to type '[boolean, boolean]'.
  Target requires 2 element(s) but source may have fewer.

fabiankaegy: Ahh that is because you manually typed the return of the function. We can leave that off. It can get inferred :)

Sidsector9: I've still moved that logic here [7c5ee73](https://github.com/10up/block-components/pull/297/commits/7c5ee73dcb61d56d3d1490db9889b7b8921a8eff) for clarity.

hooks/use-is-plugin-active/index.ts
```diff
@@ -0,0 +1,29 @@
+import { useSelect } from '@wordpress/data';
+import { store as coreStore } from '@wordpress/core-data';
+import type { Plugin } from '@wordpress/core-data';
+
+const ACTIVE_STATUSES = ['active', 'network-active'] as const;
```
Sidsector9, line 5: @fabiankaegy just wanted to confirm if `network-active` is a valid value.

It is not in the list here:

https://github.com/WordPress/gutenberg/blob/8c77df8c1052015f26b57d65fe5f043cd4f774ac/packages/core-data/src/entity-types/plugin.ts#L76

fabiankaegy: I found out about this the hard way on a project when it didn't work. I was also surprised to find it but apparently it exists.

hooks/use-is-plugin-active/index.ts
fabiankaegy: Yeah, all the Data API is a bane of my existence when it comes to TS

fabiankaegy (approved):

PR #298, opened 2024-03-01
upkeep: Add TS support for `Optional` component

Sidsector9:
### Description of the Change

Closes #

### Changelog Entry

> Changed - Add TS support to`Optional` component

### Credits

### Checklist:

### DIFF
--- a/components/optional/index.ts
+++ b/components/optional/index.ts
@@ -2,11 +2,14 @@ import PropTypes from 'prop-types';
 import { useBlockEditContext } from '@wordpress/block-editor';
 
 interface OptionalProps {
-	value?: string|number|boolean;
+	value?: string | number | boolean;
 	children: React.ReactNode;
 }
 
-export const Optional: React.FC<OptionalProps> = ({ value, children }): boolean | React.ReactNode => {
+export const Optional: React.FC<OptionalProps> = ({
+	value,
+	children,
+}): boolean | React.ReactNode => {
 	const { isSelected } = useBlockEditContext();
 	return (isSelected || !!value) && children;
 };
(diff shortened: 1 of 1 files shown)


components/optional/index.ts
```diff
@@ -1,7 +1,15 @@
 import PropTypes from 'prop-types';
```
fabiankaegy: I actually think that as part of converting something to TypeScript it makes sense to get rid of all the PropTypes stuff.

components/optional/index.ts
```diff
@@ -1,7 +1,15 @@
...
-export const Optional = ({ value, children }) => {
+interface OptionalProps {
+	value?: string | number | boolean;
+	children: React.ReactNode;
+}
+
+export const Optional: React.FC<OptionalProps> = ({
+	value,
+	children,
+}): boolean | React.ReactNode => {
```
fabiankaegy: I'm not sure we need / want to explicitly type the return type of every function. If something can be inferred I would rather just leave it up to the compiler

components/optional/index.ts
```diff
@@ -0,0 +1,25 @@
...
+	value?: string | number | boolean;
+
+	/**
+	 * The children to render if the value is truthy.
+	 */
+	children: React.ReactNode;
+}
+
+export const Optional: React.FC<OptionalProps> = ({
+	value,
```
fabiankaegy: Lets also move the `.defaultValue` to an inline declaration here

```suggestion
	value = '',
```

components/optional/index.ts
```diff
@@ -0,0 +1,25 @@
...
+	value = '',
+	children,
+}) => {
+	const { isSelected } = useBlockEditContext();
+	return (isSelected || !!value) && children;
+};
+
+Optional.defaultProps = {
+	value: '',
+};
```
fabiankaegy: This can be removed now that we have the default value inline

fabiankaegy (approved):

PR #299, opened 2024-03-01
upkeep:  Add TS support for `IsAdmin` component

Sidsector9:
### Description of the Change

Closes #

### Changelog Entry

> Changed - Add TS support to `IsAdmin` component

### Credits

### Checklist:

### DIFF
--- a/components/is-admin/index.ts
+++ b/components/is-admin/index.ts
@@ -1,7 +1,14 @@
 import { useSelect } from '@wordpress/data';
 
 interface IsAdminProps {
+	/**
+	 * Fallback component.
+	 */
 	fallback: React.ReactNode | null;
+
+	/**
+	 * Child component.
+	 */
 	children: React.ReactNode;
 }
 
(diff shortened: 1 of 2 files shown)


components/is-admin/index.ts
```diff
@@ -8,12 +13,15 @@ import { useSelect } from '@wordpress/data';
  * fallback component via the fallback prop.
  *
  * @param {object} props react props
- * @param {*} props.fallback fallback component
- * @param {*} props.children child components
- * @returns {*}
+ * @param {React.ReactNode|null} props.fallback fallback component
+ * @param {React.ReactNode} props.children child components
+ * @returns {null|React.ReactNode}
```
fabiankaegy: When we are adding types I think we can safely remove the JSDoc comments. 

If we want to add additional context to the type we can add a comment above the type value like so:

```ts
interface IsAdminProps {
    /**
     * fallback component
     */
 	fallback: React.ReactNode | null;

    /**
     * child component
     */
 	children: React.ReactNode;
 }
```

components/is-admin/index.ts
```diff
@@ -1,19 +1,29 @@
...
- * @param {*} props.fallback fallback component
- * @param {*} props.children child components
- * @returns {*}
  */
-export const IsAdmin = ({ fallback = null, children }) => {
-	const hasAdminPermissions = useSelect(
+export const IsAdmin: React.FC<IsAdminProps> = ({
+	fallback = null,
+	children,
+}): null | React.ReactNode => {
```
fabiankaegy: I still think we can leave the return type off here

fabiankaegy (approved):

PR #301, opened 2024-03-03
upkeep:  Add TS support for `PostTitle` component

Sidsector9:
### Description of the Change

Closes #

### Screenshot

Marks invalid attributes:

Screenshot 2024-03-04 at 11 29 40 AM

### Changelog Entry

> Changed - Add TS support for `PostTitle` component.

### Credits

### Checklist:


components/post-title/index.tsx
```diff
@@ -1,11 +1,18 @@
...
+interface PostTitleProps<T extends React.ElementType> {
+	tagName?: T;
+}
+
+type PostTitlePropsWithOmit<T extends React.ElementType> = PostTitleProps<T> & Omit<React.ComponentPropsWithoutRef<T>, keyof PostTitleProps<T>>;
+
+export const PostTitle = <T extends React.ElementType = 'h1'>({
+	tagName,
+	...rest
+}: PostTitlePropsWithOmit<T> ) => {
```
fabiankaegy: Not at all critical, and I am very open to being wrong here. My personal preference over single letter type generics is having a more descriptive name. 
```suggestion
interface PostTitleProps<T extends React.ElementType> {
	tagName?: T;
}

type PostTitlePropsWithOmit<TElementType extends React.ElementType> = PostTitleProps<TElementType> & Omit<React.ComponentPropsWithoutRef<TElementType>, keyof PostTitleProps<TElementType>>;

export const PostTitle = <TElementType extends React.ElementType = 'h1'>({
	tagName,
	...rest
}: PostTitlePropsWithOmit<TElementType> ) => {
```

components/post-title/index.tsx
```diff
@@ -1,11 +1,18 @@
...
-export const PostTitle = (props) => {
-	const { tagName: TagName = 'h1', ...rest } = props;
+interface PostTitleProps<T extends React.ElementType> {
+	tagName?: T;
+}
+
+type PostTitlePropsWithOmit<T extends React.ElementType> = PostTitleProps<T> & Omit<React.ComponentPropsWithoutRef<T>, keyof PostTitleProps<T>>;
+
+export const PostTitle = <T extends React.ElementType = 'h1'>({
+	tagName,
```
fabiankaegy: I would move the default value away from PropTypes into the default inline props:

```suggestion
	tagName = 'h1',
```

components/post-title/index.tsx
```diff
@@ -37,10 +46,6 @@ export const PostTitle = (props) => {
 	);
 };
 
-PostTitle.propTypes = {
-	tagName: PropTypes.string,
-};
-
 PostTitle.defaultProps = {
 	tagName: 'h1',
```
fabiankaegy: Lets move this to the inline assignment (default value) so we can remove prop types completely

Sidsector9: @fabiankaegy this PR isn't ready yet, I'm having a hard time assigning that as a default value.

I get this error:

```
Type '"h1"' is not assignable to type 'TElementType'.
  '"h1"' is assignable to the constraint of type 'TElementType', but 'TElementType' could be instantiated with a different subtype of constraint 'ElementType<any>'.ts(2322)
```

which makes sense but unsure on how to resolve it. I'll mark this PR as draft.

components/post-title/index.tsx
```diff
@@ -37,10 +46,6 @@ export const PostTitle = (props) => {
...
 };
```
fabiankaegy: @Sidsector9 Should we not remove the Default Props here?

Sidsector9: Hey thanks for taking care of this one, I was on PTO.

PR #303, opened 2024-03-03
upkeep: add TS support for `useTaxonomy` hook

Sidsector9:
### Description of the Change

Closes #

### Changelog Entry

> Changed - add TS support for `useTaxonomy` hook

### Credits

### Checklist:


package.json
```diff
@@ -55,6 +55,7 @@
     "@emotion/styled": "^11.10.5",
     "@floating-ui/react-dom": "^1.0.1",
     "@leeoniya/ufuzzy": "^1.0.8",
+    "@wordpress/core-data": "^6.29.0",
```
Sidsector9: @fabiankaegy I've installed this so that I can import the `Taxonomy` type. Can you confirm if this is the correct way to go?

fabiankaegy: @Sidsector9 I would move it to a `devDependency` because the package won't actually get bundled. But yeah this makes sense to me :)

fabiankaegy:
@Sidsector9 I'm really surprised by the amount of changes in the package lock file. Are you by chance using a different NPM version?

Sidsector9:
@fabiankaegy I'm on NPM v8.19.4.
I just tried fixing it, can you please check?

fabiankaegy (changes requested):
It seems Git didn't pick up that you renamed the file. The old .js version still is there and is getting used

fabiankaegy (approved):

PR #304, opened 2024-03-07
Fix `useMedia()` example

Soean:
### Description of the Change
Fixes the example for the `useMedia()` hook. Now uses `media.alt_text` to show the correct alt text of the media file.

### Changelog Entry

Fixed - Fix example for `useMedia()` hook

### Credits

Props @soean

### Checklist:


fabiankaegy (approved):
Good catch :)

Issue #314, opened 2024-04-30
Migrate away from `defaultProps`


fabiankaegy:
prop types / `defaultProps` were deprecated all the way in 2017 and are now getting removed in React 19. 

WordPress will likely update it's React version relatively soon after the release so we should ensure we are finally finishing the migration away from `defaultProps`

https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-deprecated-react-apis

001123:
is any tool for migrate defaultProps ?

Issue #295, opened 2024-02-27
Convert existing components to TypeScript


fabiankaegy:
A while ago we've started converting the codebase to using TypeScript #267 

Now we actually need to convert the existing components to using it. 

```[tasklist]
### API's
- [x] [registerBlockExtension](https://github.com/10up/block-components/blob/develop/api/register-block-extension)
- [x] [registerIcons](https://github.com/10up/block-components/blob/develop/api/register-icons)
```

```[tasklist]
### Components
- [x] [ClipboardButton](https://github.com/10up/block-components/blob/develop/components/clipboard-button)
- [x] [ColorSettings](https://github.com/10up/block-components/blob/develop/components/color-settings)
- [x] [ContentPicker](https://github.com/10up/block-components/blob/develop/components/content-picker)
- [x] [ContentSearch](https://github.com/10up/block-components/blob/develop/components/content-search)
- [x] [CustomBlockAppender](https://github.com/10up/block-components/blob/develop/components/custom-block-appender)
- [x] [IconPicker](https://github.com/10up/block-components/blob/develop/components/icon-picker)
- [x] [Image](https://github.com/10up/block-components/blob/develop/components/image)
- [x] [Link](https://github.com/10up/block-components/blob/develop/components/link)
- [ ] [InnerBlockSlider DEPRECATED](https://github.com/10up/block-components/blob/develop/components/inner-block-slider)
- [x] [IsAdmin](https://github.com/10up/block-components/blob/develop/components/is-admin)
- [x] [MediaToolbar](https://github.com/10up/block-components/blob/develop/components/media-toolbar)
- [x] [Optional](https://github.com/10up/block-components/blob/develop/components/optional)
- [x] [Repeater](https://github.com/10up/block-components/blob/develop/components/repeater)
- [x] [RichTextCharacterLimit](https://github.com/10up/block-components/blob/develop/components/rich-text-character-limit)
- [x] [PostAuthor](https://github.com/10up/block-components/blob/develop/components/post-author)
- [x] [PostCategoryList](https://github.com/10up/block-components/blob/develop/components/post-category-list)
- [x] [PostContext](https://github.com/10up/block-components/blob/develop/components/post-context)
- [x] [PostDate](https://github.com/10up/block-components/blob/develop/components/post-date)
- [x] [PostExcerpt](https://github.com/10up/block-components/blob/develop/components/post-excerpt)
[...]

PR #320, opened 2024-05-11
Migrate `ColorSetting` component to TS

Souptik2001:
### Description of the Change

Migrate the `ColorSetting` component to TS.

Related issue - https://github.com/10up/block-components/issues/295 and also #314 

### How to test the Change

- I have added a new block named "Color Setting Component Example", which can be used to test the functionality of the component.

### Checklist:


components/color-settings/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+	/**
+	 * Callback called when a color is selected.
+	 */
+	onChange: Function,
+
+	/**
+	 * Rest of the properties.
+	 */
+	[key: string]: any;
```
Antonio-Laguna: This is a bit arbitrary but there's no better way since we can't import the prop types for this component.

IMO this should be as `unknown` instead of `any`. From a look at `BaseControlProps` only missing here would be `id?`, `children` and `__nextHasNoMarginBottom`. Could implementing these (without the __next one) be worth it?

example/src/blocks/color-settings-example/block.json
```diff
@@ -0,0 +1,18 @@
...
+    "example": {},
+    "supports": {
+        "html": false
+    },
+    "attributes": {
+        "color": {
+            "type": "string"
+        }
+    },
+    "variations": [],
```
Antonio-Laguna: [nit] Minor indentation issue here.

example/src/blocks/color-settings-example/edit.js
```diff
@@ -0,0 +1,40 @@
...
+export const BlockEdit = (props) => {
+	const {
+		attributes: { color },
+		setAttributes
+	} = props;
+
+	const colors = [
+		{ name: 'red', color: '#f00' },
+        { name: 'white', color: '#fff' },
+        { name: 'blue', color: '#00f' },
```
Antonio-Laguna: Same issue here with indentation

Antonio-Laguna (commented):
@Souptik2001 nice work! Love the example too!

This is looking good to me. There are a couple of minor issues with indentation and I left a question. Pinging @fabiankaegy for extra pair of eyes!

components/color-settings/index.tsx
Souptik2001: Thanks @Antonio-Laguna!

I agree with your thought!

Also thinking about the three props you mentioned to use, instead of the generalized one -
- `__nextHasNoMarginBottom` - As you said it's not required. Therefore ignoring it.
- `children` - Are we supposed to pass children inside the `ColorSetting` component? If not do you think we can avoid this?
- `id?` - I see that we are explicitly setting the ID of the `BaseControl` component with a ID containing a unique Instance ID. So, should we give someone the option to override this ID (as we maybe intentionally setting it to a unique ID)?

Souptik2001:
Thanks @Antonio-Laguna!

Addressed the issues and added my thoughts on your suggestion. 👍

components/color-settings/index.tsx
Antonio-Laguna: Great points!

Then we should not be passing blindly props down to avoid any undocumented behaviour happening :) If we restrict there's no chance someone inadvertently passes down `id` and/or `children`

Souptik2001: Thanks perfect!

Removed it! 👍

Antonio-Laguna (approved):
Awesome work @Souptik2001 ! 

Thanks a lot for your contribution and your quick turnaround on my review suggestions!

fabiankaegy (approved):
Thanks for working on this :)

PR #322, opened 2024-05-14
Migrate `Counter` component to TS

Souptik2001:
### Description of the Change

Migrate the `Counter` component to TS.
It also migrates `StyledComponentContext` to TS as it was related to this and not separately listed in the issue.
Also added a `readme.md` file for this component, that was missing.

Related issue - https://github.com/10up/block-components/issues/295

### How to test the Change

- Added a new block named "Counter Example", which can be used to test the functionality of the component.

https://github.com/10up/block-components/assets/39427067/f6adbe7b-b998-44d7-8be4-9bf199d1e897

### Checklist:


components/counter/index.tsx
```diff
@@ -139,38 +147,42 @@ const CircularProgressBar = (props) => {
...
+	}, ref ) => {
+		const percentage = (count / limit) * 100;
+		return (
+			<StyledComponentContext cacheKey="tenup-component-counter">
+				<StyledCounter
+					className={cx('tenup--block-components__character-count', {
+						'is-over-limit': count > limit,
+					})}
+					ref={ref}
+				>
```
Souptik2001: Previously the rest of the props were passed down to the `StlyedCounter` component. But as this doesn't accept any props (as we can see above where it is declared using Emotion styled component), therefore I think, we don't need to pass the rest of the props down to this component.

fabiankaegy: I believe the rest is needed. The `StyledCounter` in the end is only a `styled.div` so a regular DIV which accepts any of the regular HTMLElementDiv attributes. If you for example pass an additional data attribute that should appear on the div.

Souptik2001: Yes correct, that makes sense!
Brought that back.

I got confused because of some Emotion specific stuff - https://emotion.sh/docs/typescript#react-components

components/counter/index.tsx
```diff
@@ -139,38 +147,49 @@ const CircularProgressBar = (props) => {
...
+
+	/**
+	 * Max limit.
+	 */
+	limit: number;
+
+	/**
+	 * Rest of the props.
+	 */
+	[key: string]: unknown;
```
Antonio-Laguna, line 164: ❤️ thanks for adding `unknown` this time!

components/counter/readme.md
```diff
@@ -0,0 +1,40 @@
...
+
+    return (
+		<Counter
+			count={text.length}
+			limit={20}
+		/>
+
+		<CircularProgressBar
+			percentage={(text.length / 20) * 100}
+		/>
```
Antonio-Laguna: [nit] Needs more indentation here

Antonio-Laguna (commented):
@Souptik2001 you're killing it! Thanks a lot for your contribution! Love the README and the carefulness with the types! Nice job!

I left a minor comment but this LGTM!

components/counter/index.tsx
Souptik2001: Yes, I remembered this time! 😄

components/counter/readme.md
Souptik2001: I missed wrapping it by Fragment! 🙈 
Fixed it now. 👍 Hopefully indentation is also fine now.

fabiankaegy (commented):
Love this 🚀 Thanks for your help with this :)

fabiankaegy (approved):

PR #325, opened 2024-05-15
Migrate `ContentSearch` component to TS

Souptik2001:
### Description of the Change

Migrate the `ContentSearch` component to TS.

Found two bugs in the component, which are also fixed by this PR (Please let me know if any of these are expected feature and not a bug) -
- It displays the type of the suggestion, but that doesn’t have the actual post type. Instead we should display `subtype`.
- When we click on the initial options, it doesn’t select. That’s because we have an `onBlur` event attached to the search bar, which hides the initial results whenever it's triggered. And when we try to click on any of those outside options, it just blurs out and the options just vanishes.

Both the bugs are demonstrated in the below video -

https://github.com/10up/block-components/assets/39427067/f7fe4155-b26c-4589-8f90-8ecc46c9c523

- You can see the pages show "post" as the info tag.
- When I click on any result in the initial results list, then it doesn't click and before that only the result container vanishes. This is not very clear as the selected post is not shown anywhere in the block. Therefore in this PR, I have also added a section to display the selected post.

Related issue - https://github.com/10up/block-components/issues/295 and #314 

### How to test the Change

- Added a new block named "Post Searcher Example", which can be used to test the functionality of the component.

### Checklist:


components/content-search/index.tsx
```diff
@@ -48,31 +47,87 @@ const StyledSearchControl = styled(SearchControl)`
 	width: 100%;
 `;
 
-const ContentSearch = ({
-	onSelectItem,
-	placeholder,
+interface QueryCache {
```
Souptik2001, line 50: There are many interfaces created for this component, therefore should it be better to move all the interfaces to another file say `definitions.ts`, and just import from there?

components/content-search/index.tsx
```diff
@@ -375,15 +452,12 @@ const ContentSearch = ({
...
-					onBlur={() => {
-						setIsFocused(false);
-					}}
 				/>
 
 				{hasSearchString || hasInitialResults ? (
 					<>
 						<List className={`${NAMESPACE}-list`}>
-							{isLoading && currentPage === 1 && <StyledSpinner />}
+							{isLoading && currentPage === 1 && <StyledSpinner onPointerEnterCapture={null} onPointerLeaveCapture={null} />}
```
Souptik2001, line 457: Weirdly we have to pass these two props to Spinner (specifically to SVG-based component), to stop Typescript from complaining. Otherwise, Typescript throws an error.
There is a open issue for this on the Gutenberg repo - https://github.com/WordPress/gutenberg/issues/61322

But the issue is setting these props throws warning in the console now by SVG element -

Screenshot 2024-05-16 at 1 35 12 AM

fabiankaegy (approved):
This works well in my testing :) Thanks for this enhancement!

PR #326, opened 2024-06-06
Refactor `ContentSearch` to use react query

fabiankaegy:
### Description of the Change

This PR replaces the custom `useEffect` based data fetching with custom pagination & abort controller with `react-query`

Closes #

### Changelog Entry

> Changed - Refactor `ContentSearch` component to use react-query instead of a custom data fetching logic based on useEffect

### Credits

Props @fabiankaegy @johnwatkins0

### Checklist:


components/content-search/index.tsx
```diff
@@ -514,20 +330,29 @@ const ContentSearch: React.FC<ContentSearchProps> = ({
...
 					</>
 				) : null}
 			</StyledNavigableMenu>
+	);
+};
+
+const ContentSearchWrapper: React.FC<ContentSearchProps> = (props) => {
+	return (
+		<StyledComponentContext cacheKey="tenup-component-content-search">
+			<QueryClientProvider client={queryClient}>
```
fabiankaegy, line 208: @Antonio-Laguna I'm really currious about your thoughts on this. 

Ideally the provider would wrap the entire block editor react component itself... But that isn't possible for us as extenders to do as far as I'm aware. 

So this to me seems like the next best thing?

It means that the query cache is local to each instance of the ContentSearch instead of shared for the entire page 🤔

fabiankaegy:
This work here includes improvements suggested in #221

components/content-search/index.tsx
```diff
@@ -1,13 +1,20 @@
 import { Spinner, NavigableMenu, Button, SearchControl } from '@wordpress/components';
-import apiFetch from '@wordpress/api-fetch';
-import { useState, useRef, useEffect, useCallback } from '@wordpress/element';
-import { addQueryArgs } from '@wordpress/url';
+import { useState, useRef } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import styled from '@emotion/styled';
-import SearchItem, { Suggestion } from './SearchItem';
+import {useMergeRefs} from '@wordpress/compose';
```
Antonio-Laguna: [nit] space between brackets

components/content-search/index.tsx
```diff
@@ -112,335 +98,73 @@ const ContentSearch: React.FC<ContentSearchProps> = ({
...
-					}
-				} else if (searchQuery.controller === 1 && i === currentPage) {
-					isLoading = false;
-					showLoadMore = false;
-				}
-			}
-		});
-	}
+	const searchResults = data?.pages.map((page) => page?.results).flat() || undefined;
+	console.log({searchResults});
```
Antonio-Laguna: left over

components/content-search/index.tsx
Antonio-Laguna: @fabiankaegy this looks sane to me. I don't think this is a huge issue as it stands. The chances of this being super needed because you have multiple modules being used by a single user seem very slim. Also the extra calls are not as important on the admin side so, as you say, it is the next best thing. 

Been looking to see if there was any way it could share some context but it does not seem possible (though it used to be).

components/content-search/index.tsx
fabiankaegy: Good catch :)

components/content-search/index.tsx
fabiankaegy: Something is wrong with the eslint setup... I think we need to switch over to `@typescript-eslint/parser`

fabiankaegy: Fixed the linter

PR #333, opened 2024-06-18
Fix: Refine Type Setup

fabiankaegy:
### Description of the Change

Update the remaining components to TS and refine the general type setup.


tsconfig.json
fabiankaegy, line 1: @nicholasio Would love to get your take on these changes. 

This now uses the HeadstartWP `tsconfig`

components/author/index.tsx
```diff
@@ -44,8 +44,9 @@ export const LastName: React.FC<LastNameProps> = (props) => {
 	return <TagName {...rest}>{lastName}</TagName>;
 };
 
-function useDefaultAvatar() {
+function useDefaultAvatar(): string {
 	const { avatarURL: defaultAvatarUrl } = useSelect((select) => {
+		// @ts-ignore-next-line The type definitions for the block editor store are incomplete.
```
nicholasio, line 49: use `@ts-expect-error` instead.

PR #334, opened 2024-06-25
Feature: Allow passing custom labels to `MediaToolbar` component

### LINKED ISSUE #242: MediaToolbar component: Allow configuring labels

Use case: I want to use the `MediaToolbar` with the `Video` component introduced in https://github.com/10up/block-components/pull/225, but the labels on this component assume that the bound media is an image: "Add Image", "Replace Image", "Remove Image".

https://github.com/10up/block-components/blob/6cec6c17b89abe9d19254f89b73f29ffd198a36d/components/media-toolbar/index.js#L28-L32

https://github.com/10up/block-components/blob/6cec6c17b89abe9d19254f89b73f29ffd198a36d/components/media-toolbar/index.js#L33-L37

https://github.com/10up/block-components/blob/6cec6c17b89abe9d19254f89b73f29ffd198a36d/components/media-toolbar/index.js#L44-L46

### Designs

Ideally, we'd be able to pass a `labels` argument to `MediaToolbar`, something like:

```js
<BlockControls>
	<MediaToolbar
		isOptional
		id={ video }
		onSelect={( media ) => setAttributes( {video: media.id} )}
		labels={{
			add: __( 'Add Video' ),
			replace: __( 'Replace Video' ),
			remove: __( 'Remove Video' )
		}}
	/>
</BlockControls>
```

_No response_

### Code of Conduct

Firestorm980:
### Description of the Change

Closes https://github.com/10up/block-components/issues/242

### How to test the Change

1. Check out this branch locally
2. Run the WP environment
3. Add / edit a post
4. Add the 'custom media toolbar' block
5. Confirm updated labels 

### Changelog Entry

Added - Support for custom labels in `MediaToolbar`
Fixed - Typo in documentation
Added - Custom media toolbar example

### Credits

@Firestorm980 

### Checklist:

### DIFF
--- a/components/media-toolbar/index.tsx
+++ b/components/media-toolbar/index.tsx
@@ -50,9 +50,9 @@ export const MediaToolbar: React.FC<MediaToolbarProps> = ({
 	isOptional = false,
 	id,
 	labels = {
-		replace: __('Replace Image', '10up-block-components'),
-		remove: __('Remove Image', '10up-block-components'),
 		add: __('Add Image', '10up-block-components'),
+		remove: __('Remove Image', '10up-block-components'),
+		replace: __('Replace Image', '10up-block-components'),
 	},
 }) => {
 	const hasImage = !!id;
--- a/components/media-toolbar/index.tsx
+++ b/components/media-toolbar/index.tsx
@@ -25,6 +25,15 @@ interface MediaToolbarProps {
 	 * The ID of the media, in this case the image.
 	 */
 	id: number;
+
+	/**
+	 * Labels for the buttons.
+	 */
+	labels?: {
+		replace?: string;
+		remove?: string;
+		add?: string;
+	};
 }
 
 /*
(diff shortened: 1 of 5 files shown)


components/media-toolbar/index.tsx
```diff
@@ -40,6 +49,11 @@ export const MediaToolbar: React.FC<MediaToolbarProps> = ({
 	onRemove,
 	isOptional = false,
 	id,
+	labels = {
+		add: __('Add Image', '10up-block-components'),
+		remove: __('Remove Image', '10up-block-components'),
+		replace: __('Replace Image', '10up-block-components'),
+	},
```
fabiankaegy: @Firestorm980 The downside to this way of handling the default values is that if you just want to customize the label for one of them you loose the default values of the other two. 

Also if we want to add a new label in te future we cause breakages for anyone using the custom labels because the new option would be undefined. 

Because of that I think the following would be more safe:


```js
const DEFAULT_LABELS = {
    add: __('Add Image', '10up-block-components'),
    remove: __('Remove Image', '10up-block-components'),
    replace: __('Replace Image', '10up-block-components'),
}

function ( { labels = {} } ) {

    const mergedLabels = {
        ...DEFAULT_LABELS,
        ...labels,
    };
}
```

fabiankaegy (changes requested):
Thanks for adding this enhancement :) 

Left one suggestion for how we can improve the solution. With that implemented this is ready to merge from my perspective :)

fabiankaegy (approved):
This looks good :)

fabiankaegy:
The failing test is unrelated.

PR #339, opened 2024-08-11
Correct a type declaration in the `usePrimaryTerm` component

johnbillion:
### Description of the Change

I noticed that the declared type for this entity record is incorrect. This returns a term object, not a taxonomy object.

### How to test the Change

Not sure, I only noticed this while taking a look through the component file.

### Changelog Entry

> Fixed - Fixed a declared type within the `usePrimaryTerm` component

### Credits

Props @johnbillion

### Checklist:


fabiankaegy (approved):
@johnbillion Thanks so much for your fix! :) 

Totally makes sense. Don't know how that slipped past me  👍

szepeviktor:
Maybe adding a test would prevent future slips.

PR #356, opened 2025-02-15
Fix error thrown by useRequestData hook

theskinnyghost:
### Description of the Change
The `useRequestData` hook is broken due to pulling in the store name from the `@wordpress/data` package. 
However, it looks like that package [does not export the store](https://github.com/WordPress/gutenberg/blob/1976565419e62a6924d2c73a3ec4b432701200a9/packages/data/src/index.js) as the rest of the other WP packages, hence the hook is breaking as it's trying to use it on line 30 (but not on line 24).

### How to test the Change
Create a block / editor plugin, pull in the `useRequestData` hook and try to use it.
Before these changes, you should get a JS error in the console.
After these changes, the hook should work as expected.

### Credits
Props @theskinnyghost 

### Checklist:
- [ ] I have added [Critical Flows, Test Cases, and/or End-to-End Tests](https://10up.github.io/Open-Source-Best-Practices/testing/) to cover my change.

### DIFF
--- a/hooks/use-request-data/index.ts
+++ b/hooks/use-request-data/index.ts
@@ -9,7 +9,7 @@ import isObject from 'lodash/isObject';
  */
 import { store as coreStore } from '@wordpress/core-data';
 // @ts-ignore-next-line - The type definitions for the data package are incomplete.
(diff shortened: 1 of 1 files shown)


fabiankaegy (approved):
great catch!

Issue #345, opened 2024-11-18
Using * in registerBlockExtension throws namespace validation error


palicko:
### Describe the bug

Hey,
when `*` is used in `registerBlockExtension()` to target all blocks, WP throws validation error:
```
hooks.js The namespace can only contain numbers, letters, dashes, periods, underscores and slashes.
```

### Steps to Reproduce

Register new block extension with `*`, e.g.:

```
registerBlockExtension('*', {
  extensionName: 'demo/addExtension',
  attributes: {},
  inlineStyleGenerator: () => {},
  classNameGenerator: () => {},
  order: 'before',
  Edit: () => <></>,
});
```

### Screenshots, screen recording, code snippet

![Image](https://github.com/user-attachments/assets/6fb400fa-7b64-47ef-84f3-f7ebcd6c883a)

### Environment information

- Device: Macbook
- OS: 14.6.1 
- Browser: Arc 1.69.0

### WordPress information

- WP 6.6
- PHP 8.2

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

fabiankaegy:
@palicko Thanks again for your contribution 🎉 

Version 1.19.4 was just released and contains your fix: https://github.com/10up/block-components/releases/tag/v1.19.4

PR #357, opened 2025-02-16
Fix registerBlockExtension apply condition

palicko:
Hey,
apply condition for `*` is failing because `blockName` variable was changed to `all` before `shouldApplyBlockExtension` is called.
This should fix it.

Refers [#345](https://github.com/10up/block-components/issues/345)

### How to test the Change
Using `registerBlockExtension` with `*` is working now.

### Checklist:
- [*] I agree to follow this project's [**Code of Conduct**](https://github.com/10up/.github/blob/trunk/CODE_OF_CONDUCT.md).


fabiankaegy (approved):
good catch :)

PR #358, opened 2025-02-19
Fix Media Library image preselect when using MediaToolbar

palicko:
Hey,
this fixes image preselect in Media Library when using MediaToolbar's "Replace Image".

Before:
image

After:
image

### Checklist:

- [* ] I agree to follow this project's [**Code of Conduct**](https://github.com/10up/.github/blob/trunk/CODE_OF_CONDUCT.md).


fabiankaegy (approved):
Great catch! 🚀

PR #363, opened 2025-03-14
Fix: only fetch required data in content picker

### LINKED ISSUE #362: Optimize REST API request in ContentSearch and ContentPicker components

We need to ensure that the REST API request in the ContentSearch and ContentPicker components is optimized to request only the data necessary for the component. Currently, the component requests the entire post object, which can lead to timeout issues if the object is too large. We should look into the [_fields](https://developer.wordpress.org/rest-api/using-the-rest-api/global-parameters/#_fields) parameter to see if we can trim our response size down a bit.

### Designs

_No response_

_No response_

### Code of Conduct

fabiankaegy:
### Description of the Change

This PR makes sure that we leverage the `_fields` param for fetching the already selected posts in the PostPicker to make it more performant. 

Closes #362

### Credits

Props @fabiankaegy, @claytoncollie

### DIFF
--- a/components/content-search/utils.ts
+++ b/components/content-search/utils.ts
@@ -51,6 +51,7 @@ export const prepareSearchQuery = ({
 		case 'user':
 			searchQuery = addQueryArgs('wp/v2/users', {
 				search: keyword,
+				_fields: ['id', 'link', 'type', 'name', 'subtype'],
 			});
 			break;
 		default:
@@ -61,6 +62,7 @@ export const prepareSearchQuery = ({
 				_embed: true,
 				per_page: perPage,
 				page,
+				_fields: ['id', 'url', 'type', 'title', 'subtype'],
 			});
 
 			break;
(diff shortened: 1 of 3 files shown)


components/content-search/utils.ts
```diff
@@ -61,6 +62,7 @@ export const prepareSearchQuery = ({
 				_embed: true,
 				per_page: perPage,
 				page,
+				_fields: ['id', 'url', 'type', 'title', 'subtype'],
```
claytoncollie: @fabiankaegy do we need `link` in this one? Your defaults on line 100 makes me think so.

fabiankaegy: @claytoncollie I was actually confused here... `link` didn't work for me. I needed to add `url` which then made the `link` property appear 🤔

claytoncollie: If it works, let's just go with it!

PR #366, opened 2025-03-17
Fix Typescript Builds

rleeson:
* Adjust types in the Repeater component and a few other type items


fabiankaegy (approved):
This is great! Thanks :)

PR #377, opened 2025-06-02
Add support for granular imports to reduce bundle size

### LINKED ISSUE #97: Feature: Add webpack code-splitting to package
It would be great to figure out a way to make the individual components in this package tree shakable. Right now the entire code is being imported when you import a single component and that has a major effect because you suddenly load all of the `@wordrpess/` dependencies due to the dependency extraction plgin.

jverneaut:
### Description of the Change

This PR proposes an update to the package structure to support granular imports for all components, hooks, APIs, and stores. Previously, all functionality was bundled into a single file, which led to large bundle sizes even when importing just a single component.

This PR still allows importing components, hooks, etc., from the base `@10up/block-components` entry point, but it:
- Adds an exports map in package.json for each top-level module (components, hooks, api, stores).
- Adds webpack entries for all `index.(js|jsx|ts|tsx)` files in each subfolder of those top-level modules.
- Enables granular imports, like:

```jsx
import { Image } from '@10up/block-components/components/image';
// Instead of import { Image } from '@10up/block-components';

import { registerBlockExtension } from '@10up/block-components/api/register-block-extension';
// Instead of import { registerBlockExtension } from '@10up/block-components';

import { useAllTerms } from '@10up/block-components/hooks/use-all-terms';
// Instead of import { useAllTerms } from '@10up/block-components';

import { iconStore } from '@10up/block-components/stores/icons';
// Instead of import { iconStore } from '@10up/block-components';

```

This results in bundle sizes up to 10x smaller when using granular imports instead of the global approach. The impact is especially noticeable when bundling individual Gutenberg blocks that use this library.

With these changes, the build output now looks like this:

```txt
dist/
├── api/
│   ├── register-block-extension/
│   │   ├── index.d.ts
│   │   ├── index.d.ts.map
│   │   ├── **index.js**
│   │   └── **index.js.map**
│   └── ...
├── components/
│   ├── author/
│   │   ├── index.d.ts
│   │   ├── index.d.ts.map
│   │   ├── **index.js**
│   │   └── **index.js.map**
│   └── ...
├── hooks/
│   ├── use-all-terms/
│   │   ├── index.d.ts
│   │   ├── index.d.ts.map
│   │   ├── **index.js**
│   │   └── **index.js.map**
│   └── ...
├── stores/
│   ├── icons/
│   │   ├── index.d.ts
│   │   ├── index.d.ts.map
│   │   ├── **index.js**
│   │   └── **index.js.map**
│   │   └── ...
│   └── ...
└── ...
```

### How to test the Change

I created a patch that updates the examples to use the newly added import syntax. Apply it with:
```txt
curl https://github.com/10up/block-components/commit/9f809ee507c81e89086362c21c8a33fee6aa7877.patch | git apply -v
```

Then run the tests as usual (don’t forget to reinstall packages in both the root and examples folders for the `exports` declarations to take effect).

I wasn’t able to run the Docker-based test suite locally and haven’t investigated it yet, but I was able to build the examples without issues and drop the blocks into an existing WordPress project to verify that they worked as expected (they did).

### Changelog Entry

> Developer - Add support for granular imports

### Credits
@jverneaut 

### Checklist:
- [x] I have added [Critical Flows, Test Cases, and/or End-to-End Tests](https://10up.github.io/Open-Source-Best-Practices/testing/) to cover my change.
- [ ] All new and existing tests pass. **(Not the Docker based tests because I wasn't able to run them, but no functionality has changed)**


jverneaut:
The "Compressed Size" GitHub Action gives a nice overview of what has changed:
https://github.com/10up/block-components/actions/runs/15383216260/job/43277224099?pr=377#step:4:262

The bundle size _of the library_ is indeed bigger, but the bundle size of every block built with it using granular imports shrinks significantly. This also has the benefit of being way faster for webpack to compile. I was able to build the examples blocks in 4134 ms with a dist size of 1,3 Mb when applying the provided patch with granular imports vs 14139 ms and a dist size of 8,1 Mb without them.

fabiankaegy (commented):
This is great! We orignially started talking about this one all the way in 2022. So actually having a solution for it all these years later makes me super happy :) 

I think the only thing I'd love to also do in this PR before we merge is update some of the documentation :) 

At least the main readme but maybe even the individual component file readmes to make it super clear that for most instances that is the recommended approach for importing these :)

jverneaut:
I'm glad I could help!

I had actually considered updating the README files in the PR but wasn’t sure whether you wanted this import pattern to be the recommended default or more of an opt-in approach. Happy to update the docs either way—just let me know which direction you’d prefer.

fabiankaegy:
Lets just go with the main readme for now and then when it is a bit more proven / established update all the individual docs :)

jverneaut:
I'm not sure what to update on the main README, should I change the path used as an example in the installation steps or should I add a paragraph like:

## Importing individual modules

As of [vX.X.X], you can import individual modules directly to reduce bundle size.

Instead of importing from the base entry point:
```jsx
import { Image } from '@10up/block-components';
```

You can now import directly from their specific paths:
```jsx
import { Image } from '@10up/block-components/components/image';
import { useAllTerms } from '@10up/block-components/hooks/use-all-terms';
import { registerBlockExtension } from '@10up/block-components/api/register-block-extension';
import { iconStore } from '@10up/block-components/stores/icons';
```
This provides better tree-shaking and smaller bundles, especially when only using a subset of the library.

The original import paths still work and are fully supported.

fabiankaegy:
@jverneaut yeah exactly something like that is what I would love to include :)

jverneaut:
@fabiankaegy I’ve updated the main README.md to include the new import pattern. I've included a link to a future v1.20.6, feel free to tweak it as needed.

fabiankaegy (approved):
Looking great 🎉

PR #394, opened 2025-08-29
Add useRenderAppenderWithLimit hook

### LINKED ISSUE #393: Add `useRenderAppenderWithLimit` hook

There are times when we need more granular control over how many blocks can be added to a block that supports inner blocks. That’s where the `useRenderAppenderWithLimits` custom hook comes in. With a single line of code, it gives us an easy way to enforce a maximum number of inner blocks.

For example:

```jsx
import { useRenderAppenderWithLimits } from '@10up/block-components';

const myLimitedAppender = useRenderAppenderWithLimits(4);
```

We can then apply it like this:

```jsx
const useInnerBlockProps = (
  {},
  {
    ...
    renderAppender: myLimitedAppender,
  }
);
```

And that’s it!

The hook will return either false or the block appender, depending on the maximum limit you pass in. Behind the scenes, it handles the heavy lifting by fetching and counting the inner blocks for your block, then conditionally returning the appender only when allowed.

### Designs

_No response_

_No response_

### Code of Conduct

ncoetzer:
### Description of the Change

This pull request introduces a new custom React hook, `useRenderAppenderWithLimit`, which allows developers to easily enforce a maximum number of inner blocks in WordPress blocks. It also demonstrates its usage by adding an example block and updates the block components package to export the new hook.

**New Hook Implementation and Documentation**

* Added `useRenderAppenderWithLimit` hook in `hooks/use-render-appender-with-limit/index.ts`, which conditionally returns a block appender based on the number of inner blocks and a specified limit.
* Provided documentation for the hook in `hooks/use-render-appender-with-limit/readme.md`, including usage instructions and an example.

**Example Block Demonstrating the Hook**

* Created a new block, `example/inner-block-limit`, with configuration in `example/src/blocks/inner-blocks-limit/block.json` to showcase usage of the hook.
* Implemented the block edit component in `example/src/blocks/inner-blocks-limit/edit.tsx`, utilizing the hook to limit inner blocks to four.
* Registered the block in `example/src/blocks/inner-blocks-limit/index.tsx`, wiring up the edit and save functionality.

**Package Export Update**

* Exported `useRenderAppenderWithLimit` from the main hooks index for easier import throughout the codebase.

Closes https://github.com/10up/block-components/issues/393

### Changelog Entry

> Added - New Hook

### Credits

Props @ncoetzer @fabiankaegy 

### Checklist:

- [ ] I have added [Critical Flows, Test Cases, and/or End-to-End Tests](https://10up.github.io/Open-Source-Best-Practices/testing/) to cover my change.

### DIFF
--- a/hooks/use-render-appender-with-limit/index.ts
+++ b/hooks/use-render-appender-with-limit/index.ts
@@ -1,3 +1,4 @@
+import React from 'react';
 import { useSelect } from '@wordpress/data';
 import {
 	store as blockEditorStore,
@@ -5,11 +6,11 @@ import {
 	InnerBlocks,
 } from '@wordpress/block-editor';
 
-export function useRenderAppenderWithLimit(limit: number, buttonAppender = false) {
+export function useRenderAppenderWithLimit(
+	limit: number,
+	appender: React.ComponentType = InnerBlocks.DefaultBlockAppender,
+) {
 	const { clientId } = useBlockEditContext();
-	const appenderType = buttonAppender
-		? InnerBlocks.ButtonBlockAppender
-		: InnerBlocks.DefaultBlockAppender;
 
 	return useSelect(
 		(select) => {
@@ -17,7 +18,7 @@ export function useRenderAppenderWithLimit(limit: number, buttonAppender = false
 			const { innerBlocks } = select(blockEditorStore).getBlock(clientId);
 			const numberOfInnerBlocks = innerBlocks.length;
 			const shouldRenderAppender = numberOfInnerBlocks < limit;
-			return shouldRenderAppender ? appenderType : false;
+			return shouldRenderAppender ? appender : false;
 		},
 		[clientId, limit],
 	);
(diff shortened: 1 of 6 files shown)


hooks/use-render-appender-with-limit/index.ts
```diff
@@ -0,0 +1,24 @@
+import { useSelect } from '@wordpress/data';
+import {
+	store as blockEditorStore,
+	useBlockEditContext,
+	InnerBlocks,
+} from '@wordpress/block-editor';
+
+export function useRenderAppenderWithLimit(limit: number, buttonAppender = false) {
```
fabiankaegy: API wise I would prefer if we just allowed to pass in the component that should be used as the appender as the second arg. That way it's more future proof and supports custom ones

ncoetzer: @fabiankaegy Love that idea! Thanks

ncoetzer: Done 🙂

fabiankaegy (approved):
This is looking good :)

PR #397, opened 2025-09-23
Added clarification on type attribute for content items

petenelson:
Added clarification for the `type` attribute for the `content` items. While I was working on this with a custom post type, I was running into an "Item no longer exists" error and was finally able to resolve it by using the same custom post type that I was using for the `contentTypes` parameter, rather than trying to match what I was using for the `mode` parameter.


fabiankaegy (approved):
Thanks for the addition :)

PR #401, opened 2026-01-09
Fix: Raw Titles for Content Picker

### LINKED ISSUE #384: The content picker display html code as title for some edge cases
### Describe the bug

The content picker uses the `title.rendered` property from the REST-API to render the title on the `PickedItemPreview` component.
This `title.rendered` property might contain HTML code depending on which plugin/theme is used. This HTML markup is then shown as plaintext to the user.

I think we should either render that html using `dangerouslySetInnerHTML` or we should use the `title.raw` instead of the `title.rendered`.

### Steps to Reproduce

1. Register a custom filter to the `the_title()` that adds html markup to the title
2. Select a post on the post picker
3. You will see the html markup being displayed as plaintext.

Image

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

Antonio-Laguna:
### Description of the Change

Ensures that titles are stripped of HTML before being rendered. I chose this approach vs using `dangerouslySetInnerHTML` because you can pass components to take care of rendering some things, and I thought it made sense to have this be a global fix. Furthermore, that extra HTML might make sense in a rendering scenario, but probably not for a content picker.

Closes #384 

### How to test the Change

Add a filter to `plugin.php` under example such as this one:

```php
add_filter( 'the_title', function ( $title ) {
	return '<span>' . $title . '</span>';
} );
```

Launch the dev environment and insert Content Picker into the default post.

#### Before

SCR-20260109-izmz

SCR-20260109-iznz

SCR-20260109-izov

#### After

SCR-20260109-iyqx

SCR-20260109-iytg

SCR-20260109-iysl

### Changelog Entry

> Fixed – Stripped HTML from post titles before rendering in the Content Picker to prevent unintended markup and ensure consistent display.

### Checklist:

- [X] I have added [Critical Flows, Test Cases, and/or End-to-End Tests](https://10up.github.io/Open-Source-Best-Practices/testing/) to cover my change.
- [X] All new and existing tests pass.


fabiankaegy:
I'm about to merge https://github.com/10up/block-components/pull/373 which I think also solves this problem

Antonio-Laguna:
It would've been great to know this beforehand. The PR didn't mention #384 at all :/ 

That said, I still don't see that issue closed. The code doesn't account for the things I did, but I could be wrong.

I've now fixed the conflicts here

fabiankaegy (approved):
Okay yeah I see this is different :) Sorry for conflating the two. This here looks good  👍

PR #406, opened 2026-05-05
Add useMaxInnerBlocks hook

colinswinney:
## Summary

- New `useMaxInnerBlocks` hook (TypeScript) that enforces an upper bound on a block's direct `innerBlocks`. Over-limit additions are removed and a notice is fired.
- Diffs `clientId`s against a snapshot to remove only the newest extras — existing (potentially filled) children are preserved even when a duplicate lands between them (e.g. `[A, A-copy, B]` → removes the copy, keeps `[A, B]`).
- Notice is fully customizable via a `noticeOptions` passthrough to `createNotice` — supports custom icon, action buttons, snackbar vs default type, sticky/dismissible behavior, etc. Default icon is `info` from `@wordpress/icons`.

## API

```ts
useMaxInnerBlocks({
    clientId,           // required — parent block's clientId
    max,                // required — number
    message,            // required — notice message
    status = 'warning', // 'warning' | 'info' | 'success' | 'error'
    noticeOptions = {}, // forwarded to createNotice's third argument
});
```

See [`hooks/use-max-inner-blocks/readme.md`](hooks/use-max-inner-blocks/readme.md) for full usage and customization examples.

## Files

- `hooks/use-max-inner-blocks/index.tsx` — hook implementation
- `hooks/use-max-inner-blocks/readme.md` — usage and `noticeOptions` reference
- `hooks/index.ts` — barrel export
- `README.md` — added hook link
- `example/src/blocks/max-inner-blocks-example/` — TS example block (block.json + edit.tsx + index.tsx) demonstrating every customization path

## Test plan

- [x] `npm ci && npm run build` at the repo root succeeds; `dist/index.js` exports `useMaxInnerBlocks` and types are emitted at `dist/hooks/use-max-inner-blocks/index.d.ts`
- [x] `cd example && npm run build && npm run wp-env start` boots cleanly (run via root `npm ci` since the project uses npm workspaces)
- [x] In the editor, insert the **Max Inner Blocks Example** block and verify:
  - [x] Adding a child past `max` removes the new child and fires a snackbar
  - [x] Pasting multiple children at once trims to `max`, fires one notice
  - [x] Duplicating an existing child between siblings removes the duplicate, preserves the originals
  - [x] Inspector toggles work: `max` (range 1–5), notice type (snackbar/default), icon (default/custom/none), dismissible, explicit dismiss (sticky), action button
  - [x] `noticeOptions.icon` only renders on snackbar notices (default-type `<Notice>` ignores it — documented in readme)

## CI status

[...]

PR #407, opened 2026-05-06
fix: restore CI on develop (lockfile drift + workflow regressions)

colinswinney:
## Summary

Every CI run on `develop` since #405 (2026-04-01) has been red. Jest and ESLint died at `npm ci`, and four other workflows (Cypress, compressed-size, release_testing, plus two `cypress.config.js` blockers) had latent regressions that the broken install pipeline was masking. This PR fixes them all.

## Lockfile fix

The security-patch merges in #405 left `package-lock.json` partially regenerated — some resolved versions (notably `@wordpress/eslint-plugin@18.1.0`) no longer satisfied the version ranges declared by their parents, and a tree of transitive entries went missing. Every CI `npm ci` failed with:

```
npm error code EUSAGE
npm error Invalid: lock file's @wordpress/eslint-plugin@18.1.0 does not satisfy @wordpress/eslint-plugin@17.13.0
npm error Missing: @wordpress/eslint-plugin@18.1.0 from lock file
npm error Missing: cosmiconfig@7.1.0 from lock file
... (dozens more)
```

Re-resolved with `npm install --package-lock-only` on **Node 22.22.2 / npm 10.9.7** (matching the bundled-npm version CI runners use). `package.json` is unchanged. The diff is purely additive — it adds the missing transitive entries without bumping any direct dependency.

## Workflow & config fixes

Once `npm ci` succeeded, the remaining failures surfaced and are addressed here:

### Cypress workspace build (`e2e-tests.yml` + `package.json`)

The action's inline `build:` ran `npm run build --workspaces --if-present`, which only builds the `example` workspace. But `example` imports `@10up/block-components` from `file:../dist/index.js`, which doesn't exist until the root package builds first. Updated the existing `build-test-env` script to chain root build + workspaces, and pointed the workflow at it (the cypress-io action passes the `build:` value as a single argv with no shell, so chained `&&` couldn't live inline).

### Cypress config loading (`tsconfig.json`)

When Cypress loads its config, it runs ts-node against the root tsconfig. Two issues blocked it:

- TS5107 deprecations on `esModuleInterop=false` and `moduleResolution=node10` are warnings under the toolkit build but errors under ts-node. Silenced with `ignoreDeprecations: "6.0"` (the Microsoft-recommended option).
- TS5011 fired because ts-node compiles `cypress.config.js` (and its `cypress/plugins/index.js` require) in isolation; without an explicit `rootDir`, ts-node computes the common source directory from a single file and the resolved rootDir doesn't match `outDir`. Set `rootDir: "."`.

### Cypress test specs (`Link.spec.js` + `registerBlockExtension.spec.js`)

Once Cypress could actually run, two specs failed because they clicked the post-publish "View Post" snackbar without first stripping `target="_blank"` — the click opened a new tab and the test browser stayed on the editor. Follow-up assertions then silently ran against editor markup instead of frontend markup:

- `Link.spec`: the editor's RichText `<a>` has no `href` attribute, so `should('have.attr', 'href', ...)` timed out
- `registerBlockExtension.spec`: the editor has no "Edit Page" admin-bar link, so `cy.contains('Edit Page')` timed out

`Image.spec` already strips `target` before clicking. Mirroring its pattern fixes both. Verified the underlying frontend rendering is correct by publishing a Link block via wp-cli and inspecting `<a href="https://10up.com/">First Link</a>` in the rendered HTML.

### release_testing (`publish-npm.yml`)

The job runs `npm install -g npm@latest` (upgrading to npm 11) before `npm install`. npm 11's resolver re-evaluates peer entries and writes them out, dirtying the working tree before the subsequent `npm version` step fails with `Git working directory not clean`. Switched to `npm ci` — strict no-op on lockfile state.

## Verification (local)

On **Node 20.18.1** and **Node 22.22.2**:

- [x] `npm ci` — succeeds, no `EUSAGE`
- [x] `npm run build` — succeeds, `dist/index.js` produced
- [x] `npm test` — Jest passes
- [x] `npm run build-test-env` — root + example both build in correct order
- [x] Cypress's exact ts-node config loads `cypress.config.js` cleanly — verified by reproducing Cypress's `register_ts_node.js` options in a Node harness, since Cypress 13/14's binary doesn't run on macOS 26
- [x] `wp-cli` publish of a Link block → frontend HTML contains `<a href="https://10up.com/">…</a>` (confirms the Link spec failure was about the snackbar/tab issue, not product code)

## CI status on this PR

| Check | Status |
|---|---|
| Jest | ✅ pass |
| ESLint | ✅ pass |
| Analyze (actions / javascript-typescript) | ✅ pass |
| CodeQL | ✅ pass |
| Cypress | ✅ pass (35 tests) |
| release_testing | ✅ pass |
| **build (compressed-size)** | ❌ fail — **expected and self-resolving.** The action runs `npm ci` against both PR head and base. The base (develop) still has the broken lockfile until this PR merges, so the action errors on the base-side install. Once this lands on develop, future PRs see a fixed base and the check passes without further changes. |


colinswinney:
@fabiankaegy  This was a lot of back and forth to resolve failing checks but I think this should fix all.  Once this merges, I'll merge `develop` back into #406 to hopefully resolve the red checks there.  🤞

fabiankaegy (approved):
Thanks for the fixes :)

